PR #42, opened 2026-01-05
Update core.py

itsaksdude07:
Major Bug, of listed pattern in Grammar is solved


ia-abatista:
Recommend closing, @mrgoonie. Abandoned since March and conflicting; it edits .shared/.../core.py, which is now a symlink to the source of truth.

PR #45, opened 2026-01-06
Support guideline for Jetpack Compose

tusdang33:
Add guideline for Compose, including Compose Android, Compose Desktop, Compose Web, and Compose Multiplatform, covering best practices, layouts, state management, theming, and performance recommendations


cli/assets/.shared/ui-ux-pro-max/data/stacks/jetpack_compose.csv
```diff
@@ -0,0 +1,53 @@
...
+10,SideEffect,rememberUpdatedState,Avoid stale lambdas,rememberUpdatedState,Capture directly,Safe callback,Stale state,Medium,https://developer.android.com/jetpack/compose/side-effects
+11,SideEffect,DisposableEffect,Clean up resources,onDispose,No cleanup,No leak,Memory leak,High,
+12,Architecture,Unidirectional data flow,UI → VM → State,onEvent,Two-way binding,Predictable flow,Hard debug,High,https://developer.android.com/topic/architecture
+13,Architecture,No business logic in UI,Logic belongs to VM,Collect state,Call repo,Clean UI,Fat UI,High,
+14,Architecture,Expose immutable state,Expose StateFlow,asStateFlow,Mutable exposed,Safe API,State mutation,High,
+15,Lifecycle,Lifecycle-aware collect,Use collectAsStateWithLifecycle,Lifecycle aware,collectAsState,No leak,Leak,High,https://developer.android.com/jetpack/compose/lifecycle
+16,Navigation,Event-based navigation,VM emits navigation event,SharedFlow,Nav in UI,Decoupled nav,Tight coupling,High,https://developer.android.com/jetpack/compose/navigation
```
dungngminh: In my experience, I usually avoid using `SharedFlow` to send events from the VM to the View because events can be lost when there are no subscribers (typically, subscribers only subscribe/collect from start/resume to stop/pause). The standard approach is to buffer events when there are no subscribers, and the event must be delivered only once.

There are two solutions:
- Use `StateFlow`: This is currently recommended by Google, when the view handles an event, call a function in the VM to change the `StateFlow`, avoiding handling it twice. If your document url reference to Google, why don't continue to use it.

- Use a `Channel`: but note that you must listen at `Dispatchers.Main.immediate` to avoid prompt cancellation of the Channel.

tusdang33: Thanks for the review.
My mistake — this slipped my mind. I’ve updated the guideline to address it and clarified the trade-offs, preferring Channel over StateFlow to avoid navigation event replay.

tusdang33: I agree that StateFlow isn’t inherently a bad practice if handled carefully. However, in Chris Banes’ example, the comment clearly says:

> *“A flow emitting the current message to display”*

This is **UI state**, not a one-time event, so using `StateFlow` there is appropriate.

The issue appears when `StateFlow` (or `SharedFlow`, both being hot flows) is used for one-time events:

* Events can be missed when there is no active collector
* Or unintentionally re-emitted on re-subscription
* Resetting state after collection adds unnecessary boilerplate and blurs state vs event semantics

With a properly configured `Channel`:

* No missed events
* No re-emission
* Clear intent: **events are events, state is state**

That said, if the goal here is to allow flexibility in how agents consume and handle one-time events, I’m fine with adding a guideline for using `StateFlow`, as long as it comes with explicit and safe handling rules.

So Chris Banes’ approach is valid for his use case, but it shouldn’t be generalized.
For one-time events, **`Channel` is simply a cleaner and safer fit**.

dungngminh: I can still clarify the concept of One-time events, in some contexts, that will make it easier to understand. Ultimately, we still have a solution to this problem. Thank you for the lively discussion.

dungngminh: @tusdang33 Everything is LGTM for now

PR #71, opened 2026-01-13
Fix Qoder label and include .qoder in --ai all installs

Nicenonecb:
## Summary
- show the Qoder label in AI type selection
- include .qoder when installing with --ai all

## Testing
- bun run build (cli/)
- node cli/dist/index.js init --ai all --force (verified .qoder created)


viettranx:
I've already updated cli download asset from GH first then try it self assets.

PR #166, opened 2026-03-01
fix: add error context to JSON.parse in loadPlatformConfig

ashsolei:
## Problem

`loadPlatformConfig()` calls `JSON.parse(content)` on the raw contents of a platform config file. If the file is malformed (corrupted download, manual edit typo, encoding issue), `JSON.parse` throws a generic `SyntaxError` like:

```
SyntaxError: Unexpected token } in JSON at position 42
```

This gives no indication of *which* platform config file failed, making debugging difficult — especially when `loadAllPlatformConfigs()` iterates over all platforms.

## Fix

Wrap `JSON.parse` in a try/catch that re-throws with the platform name included:

```
Invalid JSON in platform config "react.json": Unexpected token } in JSON at position 42
```

**1 file changed, 5 insertions, 1 deletion.**


ashsolei:
Closing — will maintain fixes in a private mirror instead. Thank you for the review.

PR #182, opened 2026-03-13
Fix design system color mapping and restore stack search supportfix: restore design system colors and stack search

hgzzd:
## Summary
  - fix the design system color mapping to use the actual `colors.csv` headers (`Primary`, `Secondary`, `Accent`, `Background`,
  `Foreground`)
  - restore documented `--stack` support by syncing the missing stack CSV files into `src` and loading available stacks dynamically
  - add regression tests covering both source scripts and bundled CLI assets

  ## Test Plan
  - `python3 -m unittest discover -s tests -p 'test_*.py' -v`
  - `python3 src/ui-ux-pro-max/scripts/search.py "gaming app" --design-system -f markdown`
  - `python3 src/ui-ux-pro-max/scripts/search.py "App Router" --stack nextjs`
  - `python3 cli/assets/scripts/search.py "gaming app" --design-system -f markdown`


accesspath27-lgtm (changes requested):
## Code Review: Fix design system color mapping and restore stack search

### BLOCKER: Color mapping fix is incorrect

The current `colors.csv` headers on `main` are:
```
Primary (Hex), Secondary (Hex), CTA (Hex), Background (Hex), Text (Hex)
```

The PR changes the code from `best_color.get("Primary (Hex)")` to `best_color.get("Primary")` — claiming the CSV headers are `Primary`, `Secondary`, `Accent`, `Background`, `Foreground`. **But `colors.csv` is not modified in this PR.** The actual headers still contain `(Hex)` suffixes and use `CTA`/`Text`, not `Accent`/`Foreground`.

This means the PR **breaks** color mapping — every color lookup will fall back to hardcoded defaults.

### Stack discovery: Good idea, minor concerns

- **Positive:** Replacing the hardcoded `STACK_CONFIG` dict with `_discover_stack_config()` that globs `stacks/*.csv` is a solid improvement.
- **Risk:** Runs at module import time. If `STACKS_DIR` doesn't exist, it silently returns an empty dict. Consider a warning or fallback.
- **Risk:** Assumes `filepath.stem` matches the expected `--stack` CLI argument — a CSV named `React Native.csv` would produce a `React Native` key instead of `react-native`.

### Missing CLI bundle sync

12 new stack CSVs are added under `src/ui-ux-pro-max/data/stacks/` but the CLI bundle at `cli/assets/data/stacks/` also needs them. Without this, the CLI will have different stacks than the source.

### Tests

- The color test hardcodes specific hex values (`#7C3AED`, etc.) — these will break if CSV data changes. Consider testing for valid hex codes rather than exact matches.
- Tests assume both `src/` and `cli/` scripts are runnable, but only `src/` gets the new stack CSVs.

### Other

- README update listing `--stack` values is helpful
- Whitespace-only changes in `search.py` should be a separate commit
- `STACK_HELP` moved to module level is a nice touch

### Required changes

1. Fix the color mapping to match **actual** CSV headers (`Primary (Hex)`, `Secondary (Hex)`, `CTA (Hex)`, `Background (Hex)`, `Text (Hex)`), OR also modify `colors.csv` to use the new headers
2. Sync new stack CSVs to `cli/assets/data/stacks/` as well

accesspath27-lgtm (approved):
## Updated Review — Approving

My previous review was incorrect. The diff was truncated (160KB+) and I missed that `colors.csv` **is** updated in both `src/` and `cli/` locations with the new headers (`Primary`, `Secondary`, `Accent`, `Background`, `Foreground`). The CLI stacks directory is also already synced.

### Verified

- ✅ `src/ui-ux-pro-max/data/colors.csv` — headers match code (`Primary`, `Secondary`, `Accent`, `Background`, `Foreground`)
- ✅ `cli/assets/data/colors.csv` — same headers, consistent
- ✅ `design_system.py` — both copies correctly reference new headers
- ✅ `core.py` — dynamic stack discovery works for all 13 stacks
- ✅ `search.py` — `--stack` help text auto-generated from discovered stacks
- ✅ Design system search returns real colors (`#7C3AED`, `#A78BFA`, `#F43F5E`, `#0F0F23`, `#E2E8F0`), not fallback defaults
- ✅ Stack search works for `nextjs`, `react`, `html-tailwind` (tested both src and CLI)
- ✅ All regression tests pass (2/2)

### Notes

- This is a large PR (227 files, +793/-28 shown in summary but much more in reality) — the bulk is new stack CSVs and skill additions
- The `_discover_stack_config()` approach is clean and extensible
- Apologies for the false negative on the color mapping — the truncated diff was misleading

LGTM 👍

mrgoonie:
@claude review

ia-abatista:
Recommend closing, @mrgoonie. Both fixes are already on main: design_system.py reads the correct colors.csv headers, and core.py STACK_CONFIG covers all 17 stacks (synced by #355).

PR #197, opened 2026-03-20
fix: use .agents/skills path for Antigravity (align with spec)

mohamedazimal27:
#### Issue

The CLI currently installs Antigravity skills to:

```
<workspace>/.agent/skills/
```

However, according to official Antigravity documentation, the correct path is:

```
<workspace>/.agents/skills/
```

#### Evidence

* Antigravity docs specify `.agents/skills`
* CLI currently uses `.agent/skills`
* This creates a mismatch between expected and actual skill locations

#### Fix

* Updated Antigravity platform root from `.agent` → `.agents`
* Updated CLI detection label accordingly

#### Scope

This change only affects the Antigravity platform configuration and does not impact other integrations (Claude, Cursor, Gemini, etc.).

#### Testing

* Built CLI locally
* Ran `init --ai antigravity`
* Verified `.agents/` directory is created correctly in a fresh workspace

---

Happy to add backward compatibility for `.agent` if needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,4 @@ out/
 .claude/settings.local.json
 .claude/session-state/
 ui-ux-pro-max-website/*
+.agents/
--- a/cli/assets/templates/platforms/agent.json
+++ b/cli/assets/templates/platforms/agent.json
@@ -3,7 +3,7 @@
   "displayName": "Antigravity / Generic Agent",
   "installType": "full",
   "folderStructure": {
-    "root": ".agent",
+    "root": ".agents",
     "skillPath": "skills/ui-ux-pro-max",
     "filename": "SKILL.md"
   },
--- a/cli/src/types/index.ts
+++ b/cli/src/types/index.ts
@@ -48,7 +48,7 @@ export const AI_FOLDERS: Record<Exclude<AIType, 'all'>, string[]> = {
   claude: ['.claude'],
   cursor: ['.cursor', '.shared'],
   windsurf: ['.windsurf', '.shared'],
-  antigravity: ['.agent', '.shared'],
+  antigravity: ['.agents', '.agent', '.shared'],
   copilot: ['.github', '.shared'],
   kiro: ['.kiro', '.shared'],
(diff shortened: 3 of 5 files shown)


accesspath27-lgtm (changes requested):
## Review: fix: use .agents/skills path for Antigravity

**REQUEST CHANGES** — Directionally correct but incomplete. The change only updates 2 of at least 4 locations referencing `.agent`:

### Missing updates:
1. **`cli/src/utils/detect.ts` line 22** — `existsSync(join(cwd, '.agent'))` still checks for `.agent`. After this PR, the CLI would create `.agents/` but fail to detect it as an Antigravity workspace.
2. **`cli/src/types/index.ts` line 51** — The directory mapping still references `['.agent', '.shared']`
3. **`src/ui-ux-pro-max/templates/platforms/agent.json`** — The source-of-truth template still has `"root": ".agent"`. Per the repo's sync convention, `src/` is canonical and `cli/assets/` is a copy. This PR updates the copy but not the source.

### Also needed:
- **Backward compatibility** — existing `.agent/` installs will break. The detection logic should check for both `.agent` and `.agents`.

The fix direction is right, it just needs to be complete.

ia-abatista:
Recommend closing, @mrgoonie. The Antigravity .agents/ skills path is already on main (agent.json has root .agents).

viettranx:
Thank you @mohamedazimal27 for correcting the Antigravity path toward the Agent Skills standard. I revalidated this PR against the latest `main` (`7c1806c`). Current platform configuration, detection, folder mapping, and documentation consistently use `.agents/skills`, while preserving the compatibility behavior delivered through the self-contained installer work (`28fb937`) and later updates. This old five-file branch now conflicts and is fully superseded, so I am closing it with appreciation.

mohamedazimal27:
Thank you for taking the time to revalidate the PR and for the detailed feedback. I’m glad to see the .agents/skills path is now correctly handled upstream. I appreciate the review and the opportunity to contribute! 🤝

PR #248, opened 2026-04-15
Rebrand design-system skill to Andritz Design System

Xuu-an:
## Summary
- Replaced the generic design-system skill with the Andritz Precision industrial design system
- Updated frontmatter, heading, overview, and When to Use sections to reflect Andritz branding and industrial focus
- Replaced generic token examples (blue-600/#2563EB) with Andritz tokens (andritz-blue-600/#0075be, gradient component tokens)
- Added Andritz Key Colors section with light/dark theme definitions
- Added Design Philosophy to the References table
- Updated Component Spec Pattern to use 0px radius, gradient backgrounds, tonal-shift hover, and ghost-border disabled states
- Replaced hardcoded Chart.js colors with CSS variable references
- Updated Best Practices with Andritz rules: no border-radius, tonal layering over borders, surface hierarchy for theming

## Test plan
- [x] Grep for leftover generic tokens (#2563EB, blue-600, gray-50: #F9FAFB) — none found
- [x] Verified Scripts, Slide System, and Templates sections remain intact
- [x] Reviewed via simplify skill — no issues (markdown content file, no code logic)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #249, opened 2026-04-15
feat(design-system): add Andritz design philosophy reference

Xuu-an:
## Summary
- Add "The Monolithic Engine" design philosophy document as the creative north star for the Andritz Precision design system
- Covers 10 sections: creative north star, no-line rule, glass & gradient rule, tonal architecture (light/dark surface hierarchies), typography blueprint scale, elevation & depth, data zone philosophy, industrial aesthetic rules, do's, and don'ts
- Located at `.claude/skills/design-system/references/design-philosophy.md`

## Test plan
- [ ] Verify the markdown renders correctly on GitHub
- [ ] Confirm all token values match the existing semantic/primitive token references
- [ ] Review design philosophy alignment with component specs and token architecture

### DIFF
--- a/.claude/skills/design-system/references/design-philosophy.md
+++ b/.claude/skills/design-system/references/design-philosophy.md
@@ -1,0 +1,112 @@
+# Design Philosophy: The Monolithic Engine
+
+## 1. Creative North Star
+
+The Creative North Star for the Andritz Precision design system is **"The Monolithic Engine."**
+
+This is not a standard dashboard aesthetic — it is an exercise in structural weight and industrial precision. We move away from the "floating card" aesthetic of modern SaaS toward a digital architecture that feels carved from solid carbon and steel. The experience must feel like a high-end industrial command center — authoritative, dense, and technically superior.
+
+We utilize **intentional asymmetry**: heavy data blocks anchored to a rigid left-axis while secondary metadata "floats" with generous whitespace. The contrast between technical precision and editorial layout creates a premium technical journal feel.
+
+**Key Principles:**
+- **Precision over Decoration** — Every element serves a functional purpose
+- **Structural Depth** — Hierarchy is defined by material stacking, not borders
+- **Monolithic Presence** — Large, bold typography creates an editorial "Blueprint" feel
+
+## 2. The "No-Line" Rule
+
+**EXPLICIT INSTRUCTION:** 1px solid borders for sectioning are **PROHIBITED**.
+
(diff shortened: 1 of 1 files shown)


mrgoonie:
Cron-safe maintain review: deferring this PR for now rather than merging the standalone philosophy file.

Evidence checked:
- This PR only adds `.claude/skills/design-system/references/design-philosophy.md` (+112 lines).
- Related open Andritz work already exists: #248 (rebrand SKILL.md), #251 (primitive tokens), #255 (broad Andritz design-system replacement), and #320 (new `andritz-design-system-v2` skill). #255 and #320 both contain fuller Monolithic Engine / Andritz design-system implementations.
- Current `main` does not contain an Andritz/Monolithic Engine reference, so this is not already merged, but accepting only this file would create orphaned design guidance without the matching token/component/source-of-truth changes.

Decision: deferred / needs consolidation.

Next step: please consolidate this into the chosen Andritz implementation path (prefer a narrow source-of-truth PR, or clearly supersede this with #320 if that separate skill is the intended direction). Once the owner picks the lane, this standalone PR should be retargeted or closed to avoid parallel docs drifting.

mrgoonie (changes requested):
Thanks for the contribution. I’m requesting changes because this adds a new design-system reference that is currently not wired into the skill and it introduces token names/values that do not match the existing token docs.

Required changes:
1. Link the new reference from `.claude/skills/design-system/SKILL.md` in the References table. Right now agents loading the design-system skill read the SKILL.md reference list, and this new `references/design-philosophy.md` file is not discoverable from that flow.
2. Align the token language with the current design-system token architecture, or add the missing token definitions as part of the same PR. The new doc references `surface`, `surface-container-*`, `primary`, `primary-container`, `on-surface-variant`, `outline-variant`, and `on-secondary-fixed`, but the existing semantic/primitive token docs in this repo define the current blue/gray CSS-variable model (`--color-primary`, `--color-card`, `--color-muted`, `--color-border`, etc.). As written, this creates a parallel token vocabulary that consumers cannot validate against the current design-system references.
3. Either remove the unchecked test-plan claims or include a concrete verification step. The PR body says token values should be confirmed against existing semantic/primitive token references, but the added values do not currently appear in those references.

Evidence checked:
- Diff is a single new file: `.claude/skills/design-system/references/design-philosophy.md`.
- `.claude/skills/design-system/SKILL.md` References table does not include the new file.
- Existing `references/semantic-tokens.md` and `references/primitive-tokens.md` use the repo’s current CSS-variable naming; the new Material-style surface/on-* token names are not defined there.
- Duplicate/prior-work search found related open Andritz PRs (#248, #251, #255), but no exact merged duplicate of this reference doc.

This can be a useful design direction doc, but it needs to be discoverable and token-compatible before merge.

ia-abatista:
Recommend closing, @mrgoonie. Adds an Andritz-Precision vendor design-philosophy doc to the design-system skill. Part of the off-scope Andritz series (#248-#256, #320).

PR #250, opened 2026-04-15
Replace generic states with Andritz industrial precision definitions

Xuu-an:
## Summary
- Replaced all generic state definitions in `states-and-variants.md` with Andritz-specific industrial precision patterns
- Key behaviors: tonal surface shifts for hover (not opacity), scale(0.98) active press, bottom-border-only input focus, ghost borders
- Added new sections: Hover States, Active States, Dark Theme Specific States (pulse line glow, caution/warning, ghost border dark mode)

## Test plan
- [x] Verified no leftover generic tokens (blue-500, red-600, gray-50, rounded)
- [x] All sections preserved: Interactive States, Focus, Disabled, Loading, Error, Variant Patterns, Accessibility
- [x] CSS examples use Andritz design tokens throughout

### DIFF
--- a/.claude/skills/design-system/references/states-and-variants.md
+++ b/.claude/skills/design-system/references/states-and-variants.md
@@ -1,6 +1,6 @@
-# States and Variants
+# States and Variants — Andritz Precision
 
-Component state definitions and variant patterns.
+Andritz industrial design system state definitions. Tonal shifts (not opacity) for hover, scale-[0.98] active press, bottom-border-only input focus, ghost borders.
 
 ## Interactive States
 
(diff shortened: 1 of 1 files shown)


mrgoonie (changes requested):
Thanks for the focused update. I can’t approve this as-is because the PR edits only `.claude/skills/design-system/references/states-and-variants.md`, but repo guidance says `.claude/` is local/dev output and permanent changes must come from the source-of-truth path/templates.nnBlocking issues:n- Source-of-truth mismatch: this changes a generated/local Claude skill reference directly, so the update can be lost or diverge from the canonical `src/ui-ux-pro-max/` template/data flow. Please move the durable change to the canonical source/template path that generates the design-system skill, then regenerate/sync the derived `.claude` output if needed.n- Overlap: PR #255 already proposes the broader Andritz design-system rewrite and includes this same `states-and-variants.md` change. Please either rebase this as the narrow accepted slice with the source-of-truth fix, or close it in favor of #255 if that is the intended umbrella PR.nnOnce the canonical source path is updated and the overlap is resolved, this can be reviewed again as a narrow docs/design-system change.

ia-abatista:
Recommend closing, @mrgoonie. Rebrands states-and-variants.md to Andritz Precision with hardcoded brand colors. Part of the off-scope Andritz series (#248-#256, #320).

PR #251, opened 2026-04-15
Replace primitive tokens with Andritz industrial design values

Xuu-an:
## Summary
- Replaced all generic primitive tokens (Tailwind-based gray/blue scales, generic shadows, rounded border radii) with Andritz-specific industrial precision values
- Added Andritz Blue 10-step scale, light/dark surface scales, neutral scale, and semantic status colors (error, warning, success, info)
- Switched to 8px grid spacing with extreme whitespace philosophy, blueprint typography scale with Inter/Gilroy font families, industrial 0px border radius, and ambient-only shadows
- Added dark theme support with dedicated surface and shadow tokens, glow effects, and ghost border fallbacks

## Test plan
- [x] Verified no leftover generic tokens (#2563EB, gray-50: #F9FAFB, blue-600: #2563EB)
- [x] Confirmed all seven sections present: Color Scales, Spacing, Typography, Border Radius, Shadows, Motion, Z-Index
- [x] Z-Index scale preserved as-is (not brand-specific)

### DIFF
--- a/.claude/skills/design-system/references/primitive-tokens.md
+++ b/.claude/skills/design-system/references/primitive-tokens.md
@@ -1,41 +1,65 @@
-# Primitive Tokens
+# Primitive Tokens — Andritz Precision
 
-Raw design values - foundation of the design system.
+Raw design values for industrial precision design system with dark/light themes.
 
 ## Color Scales
 
-### Gray Scale
+### Andritz Blue Scale
 
 ```css
 :root {
-  --color-gray-50:  #F9FAFB;
-  --color-gray-100: #F3F4F6;
-  --color-gray-200: #E5E7EB;
-  --color-gray-300: #D1D5DB;
-  --color-gray-400: #9CA3AF;
-  --color-gray-500: #6B7280;
-  --color-gray-600: #4B5563;
-  --color-gray-700: #374151;
-  --color-gray-800: #1F2937;
-  --color-gray-900: #111827;
-  --color-gray-950: #030712;
+  --andritz-blue-900: #001c38;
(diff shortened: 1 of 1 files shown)


mrgoonie (changes requested):
Thanks for the contribution. I’m requesting changes because this replaces the global primitive token reference with one vendor-specific Andritz system instead of adding a scoped example/preset.nnBlocking issues:n- Product fit: `design-system/references/primitive-tokens.md` is the generic baseline for the design-system skill. Replacing the gray/blue/status/spacing/type/radius/shadow primitives with Andritz-only values would make every generated design system inherit a single corporate aesthetic, which is not appropriate for a general UI/UX skill.n- Overlap: there are already broader open Andritz tracks (#255, #256, #320). This PR is only one slice of that same direction, so it should either be folded into the chosen Andritz-specific preset PR or retargeted as a scoped `andritz` example/reference instead of changing the default primitive tokens.n- Repo structure risk: the changed path is under `.claude/skills/...`. Repo docs warn `.claude/` is local/dev output and source-of-truth work should be under `src/ui-ux-pro-max/` or a clearly scoped skill/reference path. If this reference is intentionally canonical, please document why; otherwise move the change to the proper source-of-truth location.nnSuggested path: keep `primitive-tokens.md` generic, add a separate Andritz preset/reference folder or narrow this PR to documentation that shows how to derive an Andritz design system from the generic primitives.

ia-abatista:
Recommend closing, @mrgoonie. Replaces the general primitive tokens with Andritz-specific values. Part of the off-scope Andritz series (#248-#256, #320).

PR #288, opened 2026-04-26
fix: add SHA-256 checksum verification to GitHub release downloads

xiaolai:

## Security Fix (Medium)

`cli/src/utils/github.ts` — `downloadRelease()` fetches a ZIP from GitHub releases and writes it directly to disk without any integrity check. `init.ts` then extracts and installs the contents into the user's project. A compromised release asset (e.g. from a hijacked GitHub token or a compromised release pipeline) would be silently installed with no way for the user to detect the tamper.

**Fix:** This PR adds opt-in SHA-256 verification using a companion checksum asset pattern:

1. `getChecksumUrl(release, assetName)` — looks for a `<asset-name>.sha256` or `checksums.txt` asset in the GitHub release
2. `downloadRelease(url, dest, checksumUrl?)` — if `checksumUrl` is provided and the fetch succeeds, computes `SHA-256` of the downloaded buffer and compares against the expected hash before writing to disk; throws `GitHubChecksumError` on mismatch
3. `init.ts` now calls `getChecksumUrl(release, assetName)` and passes the result to `downloadRelease`

**This is backward-compatible:** if no companion checksum asset is uploaded, the behavior is identical to the current code. To enable full verification, upload a `release.zip.sha256` (or `checksums.txt`) companion file to each GitHub release — the CLI will then automatically verify before installation.

### DIFF
--- a/cli/src/commands/init.ts
+++ b/cli/src/commands/init.ts
@@ -12,9 +12,11 @@ import { logger } from '../utils/logger.js';
 import {
   getLatestRelease,
   getAssetUrl,
+  getChecksumUrl,
   downloadRelease,
   GitHubRateLimitError,
   GitHubDownloadError,
+  GitHubChecksumError,
 } from '../utils/github.js';
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
@@ -52,8 +54,10 @@ async function tryGitHubInstall(
     spinner.text = `Downloading ${release.tag_name}...`;
     tempDir = await createTempDir();
     const zipPath = join(tempDir, 'release.zip');
+    const assetName = assetUrl.split('/').pop() ?? 'release.zip';
+    const checksumUrl = getChecksumUrl(release, assetName) ?? undefined;
 
-    await downloadRelease(assetUrl, zipPath);
+    await downloadRelease(assetUrl, zipPath, checksumUrl);
 
     spinner.text = 'Extracting and installing files...';
     const { copiedFolders, tempDir: extractedTempDir } = await installFromZip(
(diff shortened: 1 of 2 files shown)


mrgoonie (changes requested):
Summary: Optional release checksum verification is a good security direction, but this PR duplicates an adjacent open checksum PR and the implementation has an unsafe fallback that silently ignores checksum fetch failures.

Risk level: Medium

Mandatory gates:
- Duplicate/prior implementation: overlap found — PR #285 implements the same SHA-256 release download verification path against the same files, and PR #315 also includes broader release/download hardening.
- Project standards: issue found — the CLI currently treats GitHub download failures as install blockers or explicit fallback paths; checksum verification should not look enabled while silently skipping a failed checksum fetch.
- Strategic necessity: clear value — verified release assets would reduce supply-chain risk for the legacy GitHub install path.

Findings:
- Important: This materially duplicates PR #285. Both PRs modify \`cli/src/commands/init.ts\` and \`cli/src/utils/github.ts\` to add optional SHA-256 verification for release downloads. Please consolidate around one checksum design instead of merging competing implementations.
- Important: When a checksum asset exists but fetching it fails, \`downloadRelease()\` currently proceeds without verification because it only checks \`if (checksumResponse.ok)\` and otherwise falls through to \`writeFile()\`. That creates a dangerous false sense of integrity protection: a transient or blocked checksum fetch silently downgrades security. If a checksum URL was selected, failure to fetch or parse it should fail closed with a clear checksum/download error.

Verdict: REQUEST_CHANGES

PR #290, opened 2026-04-27
chore: update Bun lockfile metadata

Zai69:
## Summary
- Regenerate `cli/bun.lock` using the current Bun version.
- Keep dependency installs reproducible across local environments.

## Test plan
- [x] Run `bun install` in `cli/`
- [x] Run `bun run build` in `cli/`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/cli/bun.lock
+++ b/cli/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "uxpro-cli",


Zai69:
Lockfile-only maintenance update: this change comes from running bun install with the current Bun version, which adds metadata in cli/bun.lock to keep installs reproducible across environments.

Zai69:
Ready for maintainer merge when convenient. This is a lockfile-only Bun metadata refresh () after dependency installation/build verification.

Zai69:
Correction: lockfile-only Bun metadata refresh in cli/bun.lock after dependency installation/build verification. Ready for maintainer merge.

PR #323, opened 2026-05-31
fix(cli): correct misleading docs and remove nonexistent --version hint

alfredo-petri:
## Problem

Two pieces of documentation describe behavior that doesn't match the code:

**1. `cli/README.md` — wrong install mechanism**

> "By default, `uipro init` tries to download the latest release from GitHub..."

The current default path (`init.ts:179-183`) uses `generatePlatformFiles()` — template generation from bundled assets. GitHub download only happens in the legacy ZIP mode (`init.ts:160`), which is not exposed as a CLI option. So `--offline` is currently a no-op relative to the default.

**2. `cli/src/commands/versions.ts:34` — advertises a nonexistent flag**

```
Use: uipro init --version <tag> to install a specific version
```

`--version` is not a registered option in `index.ts` for the `init` command. Running this would either use Commander's built-in `--version` (prints CLI version) or fail silently.

## Fix

- Rewrite "How It Works" in README to accurately describe template-based generation
- Replace the nonexistent `--version` hint in `versions.ts` with an accurate tip

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -38,9 +38,9 @@ uipro update                # Update to latest version
 
 ## How It Works
 
-By default, `uipro init` tries to download the latest release from GitHub to ensure you get the most up-to-date version. If the download fails (network error, rate limit), it automatically falls back to the bundled assets included in the CLI package.
+`uipro init` generates skill files from templates bundled inside the CLI package. This means installation works offline by default and does not require a network connection.
 
-Use `--offline` to skip the GitHub download and use bundled assets directly.
+Use `--offline` to make the offline behavior explicit (no-op with the current default, kept for clarity and future compatibility).
 
 ## Development
 
--- a/cli/src/commands/versions.ts
+++ b/cli/src/commands/versions.ts
@@ -31,7 +31,7 @@ export async function versionsCommand(): Promise<void> {
     });
 
     console.log();
-    logger.dim('Use: uipro init --version <tag> to install a specific version');
+    logger.dim('Run uipro update to reinstall the skill with the current bundled version.');
   } catch (error) {
     spinner.fail('Failed to fetch versions');
     if (error instanceof Error) {


mrgoonie (changes requested):
Summary: This PR correctly removes the nonexistent `uipro init --version <tag>` hint, but the README rewrite goes too far and now makes `--offline` look like a no-op even though it still changes behavior in the legacy ZIP install path.

Risk level: Medium

Mandatory gates:
- Duplicate/prior implementation: no merged duplicate found; search found this PR plus broader CLI/release PRs, but no prior fix for the exact `--version` hint + README wording.
- Project standards: issue found — CLI docs should describe both default template mode and the remaining legacy/offline behavior accurately.
- Strategic necessity: clear value — avoiding nonexistent CLI flags and stale install docs reduces support churn.

Findings:
- Important: `cli/src/commands/init.ts` still has a legacy ZIP-based install branch where `options.offline` skips the GitHub release download and falls back to bundled assets (`if (options.legacy) { ... if (!options.offline) tryGitHubInstall(...) ... copyFolders(...) }`). The PR README says `--offline` is a no-op with the current default, but the public CLI no longer exposes `--legacy` in `cli/src/index.ts`, so this wording is confusing from both directions: it documents internal behavior without explaining that the GitHub-download path is legacy-only/inaccessible from current public options. Please adjust the README/help text so it is precise, e.g. default `init` uses bundled template generation; `--offline` is retained for compatibility and only affects legacy/package paths if those are exposed, or remove/clarify the flag separately.

Verdict: REQUEST_CHANGES

viettranx:
Thank you @alfredo-petri for catching both misleading CLI messages. I revalidated them against the latest `main` (`7c1806c`). `cli/README.md` now states that `uipro init` generates from bundled templates, describes `--offline` as a compatibility flag, and tells users to update the package before regenerating. `versions.ts` no longer advertises the nonexistent `uipro init --version` flag; it gives the current update-and-init guidance instead. These corrections landed with the focused CLI asset/release work in #355 (`b4d33be`) and later updates. Since this old branch conflicts and is fully superseded, I am closing it with appreciation.

PR #324, opened 2026-05-31
fix(cli): implement --force flag to protect existing skill files

alfredo-petri:
## Problem

`--force` is registered in `index.ts` and described in the README, but has no effect. `template.ts` calls `writeFile()` unconditionally without checking whether the file already exists.

This means:
- Users running `uipro init` a second time silently overwrite their customized skill files
- The `--force` flag advertised in the README is misleading — it does nothing

## Fix

Add a file existence check in `generatePlatformFiles` before calling `writeFile`:

- If file exists and `--force` was not passed → skip with a clear message
- If file exists and `--force` was passed → overwrite (previous behavior for all cases)
- Propagate `force` through `generateAllPlatformFiles` → `templateInstall` → `initCommand`

## Behavior after this PR

```bash
uipro init --ai claude          # skips if SKILL.md already exists, prints warning
uipro init --ai claude --force  # overwrites regardless
```

### DIFF
--- a/cli/src/commands/init.ts
+++ b/cli/src/commands/init.ts
@@ -101,17 +101,18 @@ async function templateInstall(
   targetDir: string,
   aiType: AIType,
   spinner: ReturnType<typeof ora>,
-  isGlobal = false
+  isGlobal = false,
+  force = false
 ): Promise<string[]> {
   spinner.text = isGlobal
     ? 'Generating skill files globally...'
     : 'Generating skill files from templates...';
 
   if (aiType === 'all') {
-    return generateAllPlatformFiles(targetDir, isGlobal);
+    return generateAllPlatformFiles(targetDir, isGlobal, force);
   }
 
-  return generatePlatformFiles(targetDir, aiType, isGlobal);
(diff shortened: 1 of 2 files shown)


mrgoonie (approved):
Summary: This is a focused fix for the advertised --force behavior in the template installer path. It correctly threads the force option through init -> templateInstall -> generatePlatformFiles and prevents accidental overwrite of an existing generated skill file when --force is absent.

Risk level: Low

Mandatory gates:
- Duplicate/prior implementation: clear — searched existing issues/PRs for the force/overwrite behavior and found no overlapping implementation besides this PR.
- Project standards: passed — the change stays inside the CLI init/template generation path and reuses the existing fs access helper pattern.
- Strategic necessity: clear value — avoiding silent overwrites protects user-customized skill files and makes the documented --force flag meaningful.
- CI/checks: missing/unstable — no status checks are reported for this PR in GitHub, and this cron environment does not have bun installed to run the package build.

Findings:
- Critical: none.
- Important: none.
- Suggestion: consider adding a small CLI/template unit or smoke test later for the no-force skip path and the --force overwrite path, because this behavior is easy to regress.

Verdict: APPROVE

PR #326, opened 2026-05-31
fix(cli): make uipro update actually upgrade the CLI via npm

alfredo-petri:
## Problem

`updateCommand` (`cli/src/commands/update.ts:25`) called `initCommand({force: true})` after showing the latest GitHub release tag. Since `initCommand` uses locally bundled assets by default, the actual CLI binary was never updated — only the skill files were reinstalled from the version already on disk.

A user running `uipro update` expecting a version bump received the same files they already had.

## Fix

Replace the body of `updateCommand` with a call to `npm install -g uipro-cli@latest` via `execSync`. This correctly upgrades both the CLI binary and its bundled assets.

- Still fetches latest release info to show the version (best-effort, non-blocking)
- Clear error message if npm fails
- Post-install guidance: `uipro init --ai <type> --force` to reinstall the skill
- Remove `--ai` option from `update` (was forwarded to initCommand which no longer drives update)

## Before / After

```bash
# Before: reinstalled same bundled assets, no CLI upgrade
uipro update

# After: runs npm install -g uipro-cli@latest
uipro update
```

### DIFF
--- a/cli/src/commands/update.ts
+++ b/cli/src/commands/update.ts
@@ -1,33 +1,36 @@
+import { execSync } from 'node:child_process';
 import chalk from 'chalk';
 import ora from 'ora';
 import { getLatestRelease } from '../utils/github.js';
 import { logger } from '../utils/logger.js';
-import { initCommand } from './init.js';
-import type { AIType } from '../types/index.js';
 
-interface UpdateOptions {
-  ai?: AIType;
-}
-
-export async function updateCommand(options: UpdateOptions): Promise<void> {
+export async function updateCommand(): Promise<void> {
   logger.title('UI/UX Pro Max Updater');
 
-  const spinner = ora('Checking for updates...').start();
+  const spinner = ora('Checking latest version...').start();
 
   try {
     const release = await getLatestRelease();
-    spinner.succeed(`Latest version: ${chalk.cyan(release.tag_name)}`);
+    const latest = release.tag_name;
+    spinner.succeed(`Latest release: ${chalk.cyan(latest)}`);
+  } catch {
+    spinner.warn('Could not fetch latest release info. Proceeding with update anyway.');
+  }
 
-    console.log();
-    logger.info('Running update (same as init with latest version)...');
-    console.log();
+  console.log();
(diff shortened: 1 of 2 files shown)


mrgoonie (changes requested):
Summary: This PR fixes the right stale-update symptom, but the update command now shells out to a globally resolved npm binary from inside the installed CLI. That is not safe enough for a CLI maintainer path.

Risk level: Medium

Mandatory gates:
- Duplicate/prior implementation: clear — no merged PR already fixes `uipro update`; related release/package issues remain open.
- Project standards: issue found — CLI command execution should avoid shell string expansion when arguments are static.
- Strategic necessity: clear value — users expect `uipro update` to update the actual CLI package, not reinstall bundled assets from the old binary.
- CI/checks: locally blocked — `npm --prefix cli run build` requires `bun`, which is not available in this cron environment.

Findings:
- Important: `cli/src/commands/update.ts` uses `execSync('npm install -g uipro-cli@latest', { stdio: 'inherit' })`. Even though the current command string is static, this routes through a shell and executes whichever `npm` is first on PATH from a CLI command that users may run in arbitrary projects. Please switch to `execFileSync` / `spawnSync` with explicit args (`npm`, `["install", "-g", "uipro-cli@latest"]`) and keep the error handling, so the updater does not add an unnecessary shell-execution surface.
- Suggestion: update the README/CLI README command description to make it clear that `uipro update` updates the global CLI and users should rerun `uipro init --ai <type> --force` after it.

Verdict: REQUEST_CHANGES

ia-abatista:
Recommend closing, @mrgoonie. Superseded by merged #389, which makes uipro update actually upgrade via npm using execFileSync (no shell).

mrgoonie:
Cron maintainer follow-up: closing this as superseded by merged PR #389, which implements the same `uipro update` behavior using `execFileSync` with explicit npm args and also updates the CLI README guidance. This resolves the previous shell-execution blocker on this PR and avoids keeping two competing update-command implementations open.

mrgoonie:
Closed by maintainer automation: superseded by merged #389.

Issue #353, opened 2026-06-12
app-interface.csv shows newer repo/release state version 2.5.0, but npm still ships older version 2.2.3


ia-abatista:
Hey guys, when running the 'uipro init --ai codex' or even 'uipro update' the previous version skills are installed and not the new version.

ia-abatista:
I submitted a fix, please review and approve is appropriate: https://github.com/nextlevelbuilder/ui-ux-pro-max-skill/pull/355

Not sure if @Spinulosa1111 addressed these in his unpublished fixes

Spinulosa1111:
Hi @ia-abatista — yes, this is exactly what the unpublished fixes address. The root cause is that the repo has never had an automated release pipeline, so npm has been stuck on 2.2.3 while the repo moved to 2.5.0.

PR #348 includes a GitHub Actions release workflow that:
- Resolves symlinks before packaging (fixing the install failures too)
- Builds and publishes a clean ZIP + `checksums.sha256` automatically when a version tag is pushed
- Bumps the npm package to match the repo version

Once @nextlevelbuilder merges the queued PRs and tags `v2.5.1`, the workflow will publish automatically and this gap will be closed for good.

Your PR #355 may overlap with some of this — @nextlevelbuilder should review both together to avoid duplication. Happy to coordinate if needed!

ia-abatista:
Thanks for responding @Spinulosa1111 - I really hope @nextlevelbuilder responds soon, the repo has been stale for several months now, would be a shame to abandon it. The community should probably consider forking or hopefully others can get additional access to help manage it.

Ray0907:
Hi @ia-abatista @Spinulosa1111, I’m one of the contributors here, but I’m not a maintainer and I don’t have merge or release access to this repository.

Since upstream has been quiet for a while and this issue affects the published CLI/package flow, I’ve started a clearly labeled community-maintained fork here:

https://github.com/Ray0907/ui-ux-pro-max-skill

The goal is not to replace the original project or present the fork as official. It is just to keep fixes, releases, and package updates moving while maintainer access is unavailable. If @nextlevelbuilder resumes maintenance or grants access, I’m happy to help upstream the changes and move work back here.

I’ll keep the original attribution and MIT license intact, and any community build will use a separate package name so it does not conflict with the official `uipro-cli` package.

The first areas I’d like to focus on are the issues already discussed here:

- npm/release version mismatch
- CLI asset sync
[...]

PR #322, opened 2026-05-31
fix(cli): sync cli/assets with src/ui-ux-pro-max source of truth

alfredo-petri:
## Problem

`cli/assets/` was divergent from `src/ui-ux-pro-max/` (the documented source of truth per CLAUDE.md). Users installing via npm received outdated data:

- `scripts/core.py`: only 3 stacks instead of 16
- `scripts/search.py`: stale domain/stack documentation
- 12 stack CSVs out of date
- 18 platform JSON templates with wrong stack counts/descriptions
- Several data CSVs divergent (app-interface, colors, icons, landing, products)

**Root cause:** The sync step (`cp -r src/ui-ux-pro-max/...`) is documented in CLAUDE.md but was never automated, so it drifts with every commit to `src/`.

## Solution

- Add `cli/scripts/sync-assets.js` — a zero-dependency Node.js script that copies `src/ui-ux-pro-max/{data,scripts,templates}` → `cli/assets/`
- Add `sync-assets` npm script for manual runs
- Wire `prepack` and `prepublishOnly` to run sync automatically before every `npm pack` / `npm publish`
- Run sync to fix current 33-file divergence

## Verification

```bash
node cli/scripts/sync-assets.js
diff -rq src/ui-ux-pro-max/data cli/assets/data      # empty
diff -rq src/ui-ux-pro-max/scripts cli/assets/scripts  # empty
diff -rq src/ui-ux-pro-max/templates cli/assets/templates # empty
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #355, opened 2026-06-12
fix CLI asset sync

ia-abatista:
## Summary

This PR fixes stale CLI install payloads so `uipro init --ai *` installs the current UI/UX Pro Max skill content.

## Root Cause

`uipro init` generates files from bundled `cli/assets`, but those assets had drifted from the canonical `src/ui-ux-pro-max` source tree. The uipro cli installs were therefore missing current stack support and metadata even though the source skill had been updated.

## Changes

- Sync `cli/assets` from `src/ui-ux-pro-max` so Codex and other CLI installs get current data, scripts, and templates.
- Add `npm run sync:assets` and `npm run check:assets` using a repo-owned sync/check script.
- Run asset sync and TypeScript checks from `prepublishOnly` to prevent stale npm payloads.
- Fix the TypeScript `all` vs concrete AI platform type mismatch in uninstall detection.
- Update CLI docs and update/version messaging so the CLI no longer claims unsupported latest-version behavior.

## Validation

- `npm run check:assets`
- `npm run typecheck`
- `python cli/assets/scripts/search.py "server components" --stack nextjs`
- Temp `uipro init --ai codex`, then installed search script returned Next.js stack guidance.


ia-abatista:
Hey @nextlevelbuilder I submitted a patch to fix the current cli issue. Please review and merge to main when possible. This addresses issue mentioned here: https://github.com/nextlevelbuilder/ui-ux-pro-max-skill/issues/353

mrgoonie (changes requested):
Summary: This targets the right packaging problem, but the current PR is not merge-ready because its own asset drift gate fails after checkout.nnRisk level: MediumnnMandatory gates:n- Duplicate/prior implementation: overlap found with #322, #330, #336; this PR is the most complete automation path but should be reconciled with those before merge.n- Project standards: issue found. The PR adds `npm run check:assets`, but that check currently fails on this branch.n- Strategic necessity: clear value. Fixing stale CLI assets directly addresses the npm/CLI packaging gap discussed in #353 and #362.n- CI/checks: missing / locally failing validation.nnFindings:n- Important: `npm --prefix cli run check:assets` fails on this PR with stale asset reports for `data/products.csv` and multiple `data/stacks/*.csv` files. The byte diffs appear to be line-ending drift between `src/ui-ux-pro-max` and `cli/assets`, but because `check:assets` hashes raw bytes, this will block `prepublishOnly` and make the new release guard fail. Please normalize the synced assets or make the checker intentionally normalize line endings before hashing.n- Important: The PR overlaps materially with #322, #330, and #336. Please state why this PR supersedes or intentionally combines those approaches, especially around real-file packaging vs source-of-truth syncing.nnVerdict: REQUEST_CHANGES

ia-abatista:
Thanks for the detailed review -- addressed both mandatory gates.

**Re: `check:assets` failing** -- Fixed (`dd6dd25`). Root cause was line-ending drift: `src/` CSVs are stored LF, the `cli/assets/` copies CRLF, and `fileHash` hashed raw bytes, so `git autocrlf` on checkout produced spurious "stale" reports for `products.csv` and `data/stacks/*.csv`. The checker now normalizes `CRLF→LF` before hashing (and `sync:assets` writes LF-normalized copies), comparing content, not line endings. Reproduced the failure on a forced src=LF / asset=CRLF tree -- old checker reports 3 stale, new checker passes -- and a full `sync -> check` round-trip is green.

**Re: making it a real release guard** -- Added `.github/workflows/check-asset-sync.yml` (`e71b5f4`). It runs `npm run check:assets` on every PR touching `src/ui-ux-pro-max/**` or `cli/assets/**`, so drift hard-fails in CI. Verified locally: real content drift → exit 1; EOL-only differences -> pass. No install step (the script is pure node builtins).

**Re: overlap with #322, #330, #336** -- Different layers; this PR is the enforceable mechanism the others lack:

- **#330** -- assets-only one-time re-sync. The content it restored already lives in the source of truth (`src/ui-ux-pro-max` has all 16 stacks + a healthy `landing.csv`); `npm run sync:assets` regenerates it deterministically and `check:assets` + CI prevent re-drift. Supersedes the manual snapshot.
- **#322** -- closest overlap; also adds a sync script (`sync-assets.js`). This supersedes it: ESM `.mjs` matching the package's `"type":"module"`, a `--check` mode wired into a CI guard, and line-ending normalization so the check is platform-stable. Happy to fold in anything unique from #322.
- **#336** -- operates at the `.claude/skills` layer (real copies for Windows) and also proposed a `check-asset-sync.yml`, but with a `cli/assets` soft-fail because raw `diff` trips on the same CRLF/LF drift this PR fixes at the checker. Complementary. I named my workflow the same file so they reconcile to one guard — if #336 lands first, I'll drop mine and just wire `npm run check:assets` into theirs (no more soft-fail needed); if this lands first, #336 can drop its `cli/assets` diff step and keep its `.claude/skills` checks.

mrgoonie (approved):

Issue #334, opened 2026-06-04
Can't Install into Claude Skills


hendoscott7:
"Zip file contains a symbolic link", I'll move on and come back if I remember.

mrgoonie:
Summary: This is a real packaging/install symptom, but the report does not yet include enough installation context to identify the exact failing artifact.

Decision: needs-info / likely packaging symlink issue

Evidence:
- The error text is `Zip file contains a symbolic link`, which matches the active packaging/symlink thread already being tracked in this repo.
- Related open work includes PRs #336/#368 for replacing or packaging symlinked assets, plus broader release/package distribution issue #353 and CLI bundle issue #362.
- This issue does not include the install method, CLI/package version, platform, or the exact command/output, so I cannot distinguish Claude Code plugin ZIP install vs npm CLI bundle vs source archive install.

Next step: please paste the smallest reproducible context:

```bash
# install path you used, e.g. /plugin install URL, npm, npx, zip upload, source clone
node --version || true
npm --version || true
npx uipro-cli --version || true
# exact command or UI flow that produced: Zip file contains a symbolic link
```

With that, this can be linked to the right packaging fix instead of guessing across the symlink-related PRs.

clark-cant:
Maintainer follow-up (2026-06-26): This issue has been open for 22 days with `maintain:needs-info` and `decision:needs-repro` labels. No reproduction details have been provided since the initial request on 2026-06-21.

To move this forward, we still need:
- Exact install method (plugin install, npm, npx, zip upload, source clone)
- Platform and Node/npm versions
- Exact command or UI flow that produced the error
- Full error output or screenshot

Without these details, this will be closed as stale in approximately 5 days (around 2026-07-01) per the project's needs-info staleness policy.

The underlying symptom (`Zip file contains a symbolic link`) is a known packaging issue being tracked in PRs #336, #368 and related issues #353, #362.

*Posted by github-maintain cron at 2026-06-26T09:00:00Z*

clark-cant:
Cron maintainer stale-candidate warning: this issue has been open for 23 days (since 2026-06-04) with `maintain:needs-info` and `decision:needs-repro` labels. A follow-up request was posted on 2026-06-26 asking for install method, platform/Node versions, and exact error output — no response received.

**To keep this issue open**, please reply with:
- Install method used (plugin marketplace, npm, npx, zip upload, source clone)
- Platform (OS + version) and Node/npm versions
[...]

PR #368, opened 2026-06-19
Replace symlinks with real files and add conversion script

jofoo277:


mrgoonie:
Summary: I’m deferring this in the cron-safe maintainer lane because the PR is too large for a bounded review pass.

Decision: deferred

Evidence:
- This PR changes 37 files with +9,300/-2 lines.
- Most of the diff replaces symlinked generated skill assets under .claude/skills/ui-ux-pro-max/data and scripts with real files, plus adds scripts/convert_symlinks.sh.
- That overlaps with the existing packaging/release/version mismatch thread and related PRs, so it needs a grouped packaging review instead of a quick metadata-only approval.

Next step: please narrow this PR to either (1) the conversion script + documented generation process, or (2) the generated asset replacement, and explain how this should be reconciled with the current source-of-truth files and release workflow. A follow-up maintainer pass should compare this with the other CLI asset sync / symlink / release PRs before approving.

mrgoonie (changes requested):
Summary: This PR addresses the real symlink-install problem by replacing `.claude/skills/ui-ux-pro-max/data` and `scripts` symlinks with real files, but it is not merge-ready in its current shape.

Mandatory gates:
- Duplicate/prior implementation: overlaps the same packaging/symlink problem tracked by PR #336 and issue #334/#362, but this PR targets the `.claude/skills/ui-ux-pro-max` symlink copies specifically.
- Project standards: source-of-truth should stay under `src/ui-ux-pro-max/`; mirrored `.claude/skills/...` assets must be reproducible and kept in sync.
- Strategic necessity: the goal is valid because ZIP/plugin installs can reject symlinks, but the implementation must not introduce stale generated copies or broken helper scripts.

Requested changes:
1. Fix `scripts/convert_symlinks.sh`: `found=1` is set inside a piped `while` subshell, so the final `if [ "$found" -eq 0 ]` still sees `0` and prints `No symlinks found.` even after replacements. Use process substitution or another no-subshell loop.
2. Do not add `.claude/skills/ui-ux-pro-max/data/_sync_all.py` as a generated one-off maintenance script inside the mirrored skill data directory. It embeds product-specific synchronization logic and should either live in a normal repo `scripts/` path with docs/tests or be omitted from the packaged mirror.
3. Add a reproducible validation path for the mirror: a command/test that confirms the repo/package contains no symlinks in installable plugin/skill artifacts and that mirrored files match the `src/ui-ux-pro-max/` source-of-truth.
4. Reconcile this with the related packaging/symlink PRs so maintainers do not merge two competing mirror strategies.

Verdict: Request changes. The problem is worth fixing, but this PR currently adds a very large copied mirror plus a broken conversion script without enough reproducibility guarantees.

clark-cant:
Cron maintainer follow-up: this PR is still mergeable and the review from 2026-06-24 requested specific changes that have not been addressed yet:

1. Fix `scripts/convert_symlinks.sh` subshell variable scoping bug
2. Remove `_sync_all.py` from the mirrored skill data directory
3. Add a reproducible validation command proving mirrored files match `src/ui-ux-pro-max/` source-of-truth
4. Reconcile with overlapping packaging PRs (#346, #336)

No new commits since the review (head: `d7eeaac8`). This remains deferred until the author responds or updates the branch.

*Posted by github-maintain cron at 2026-06-26T08:50:46Z*

PR #375, opened 2026-06-22
feat(release): add semantic release automation

mrgoonie:
## Summary
- add semantic-release config for stable releases from `main` and beta prereleases from `dev`
- add GitHub Actions release workflow that installs/builds the Bun CLI, publishes `uipro-cli`, creates GitHub releases, and updates changelog/version files
- add version sync script for `skill.json`, Claude plugin metadata, marketplace metadata, and CLI manifests
- document Conventional Commit release rules and required `NPM_TOKEN`

## Release behavior
- `dev` -> beta prerelease, npm `beta` dist-tag, GitHub prerelease
- `main` -> stable release, npm `latest` dist-tag, GitHub release
- Conventional Commits drive semantic bumps and generated changelog

## Verification
- `node --check scripts/sync-release-version.mjs`
- `node -e "JSON.parse(require('fs').readFileSync('.releaserc.json','utf8'))"`
- `node scripts/sync-release-version.mjs 9.8.7-beta.1` then verified all 7 version fields changed
- restored versions to `2.5.0`
- `git diff --check`

## Notes
- Repository must define `NPM_TOKEN` with publish access to `uipro-cli` before the workflow can publish npm packages.
- Local CLI build was not run here because this runner has no `bun`; the workflow installs Bun via `oven-sh/setup-bun`.

PR #384, opened 2026-06-25
feat(release): publish uipro-cli to npm on release (#353)

ia-abatista:
## What & why

`uipro init` / `uipro update` users still receive **2.2.3-era** assets even though the repo has advanced to **v2.7.0** (issue #353). Root cause: the `Release` workflow runs semantic-release with only the **github** plugin, so it creates GitHub releases/tags but **never publishes the `uipro-cli` package to npm**.

| Source | Version |
|---|---|
| npm `uipro-cli` (what users install) | **2.2.3** |
| newest git tag | **v2.7.0** |

## Change

Add `@semantic-release/npm` to the release pipeline so every release also publishes the CLI package at the tagged version:

- **`.releaserc.json`** — add `@semantic-release/npm` with `pkgRoot: "cli"` (between `release-notes-generator` and `github`).
- **`.github/workflows/release.yml`** — add `-p @semantic-release/npm` to the `npx` invocation and `NPM_TOKEN` to the step env.

The package's existing `prepublishOnly` (`sync:assets && typecheck && build`) runs on publish, so the tarball always ships current, in-sync assets.

## Required maintainer setup

⚠️ **Add a repo secret `NPM_TOKEN`** — an npm **automation** token with publish rights to `uipro-cli`. Without it, the npm step fails at `verifyConditions`.

After this merges and a release runs, npm will receive the current version (e.g. `2.7.x`/`2.8.0`) instead of `2.2.3`.

## Scope

This is the **npm-publish slice** of the release lane mrgoonie outlined (synced assets ✅ #355 · working release workflow ✅ on main · **npm publish/tag → this PR** · safe updater command → #326). Intentionally leaves #353 open until the full distribution path lands.

## Notes / optional follow-up

- semantic-release sets the published version from the git tag; it does not commit the bump back to `cli/package.json` (stays cosmetic at 2.5.0). If you want the repo's `package.json`/`skill.json` version to track the published version, add `@semantic-release/git` in a follow-up.
- `id-token: write` is already granted, so npm provenance (`NPM_CONFIG_PROVENANCE: true`) can be enabled later if desired.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,11 +52,13 @@ jobs:
       - name: Run semantic-release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
         run: |
           npx --yes \
             -p semantic-release \
             -p @semantic-release/commit-analyzer \
             -p @semantic-release/release-notes-generator \
+            -p @semantic-release/npm \
             -p @semantic-release/github \
             -p conventional-changelog-conventionalcommits \
             semantic-release
--- a/.releaserc.json
+++ b/.releaserc.json
@@ -22,6 +22,12 @@
       }
     ],
     [
+      "@semantic-release/npm",
+      {
+        "pkgRoot": "cli"
+      }
+    ],
+    [
       "@semantic-release/github",
       {
         "successComment": false,


mrgoonie (approved):
Reviewed the npm-publish slice for #353. This is the narrow missing step after the merged semantic-release workflow: it adds @semantic-release/npm with pkgRoot=cli, passes NPM_TOKEN to the release step, and keeps the existing GitHub release behavior intact.

Checks performed:
- Compared against current release workflow/config on main and merged PR #375. #375 explicitly left npm publishing out; this PR fills that gap rather than duplicating it.
- Cross-checked issue #353 and recent release-lane comments: PR #355 covers asset sync; this PR covers the remaining npm publish/tag path.
- Verified changed JSON parses and the workflow only adds the npm plugin/env needed for semantic-release.

No blocking code findings. Maintainer setup is still required: add an npm automation token as repository credential NPM_TOKEN with publish rights for uipro-cli before the next release run, otherwise semantic-release/npm will fail verifyConditions.

PR #302, opened 2026-05-10
[codex] Add direct Codex skill package

Sabotage1:
## Summary

Adds a self-contained Codex skill package at `.codex/skills/ui-ux-pro-max`.

The existing `.claude/skills/ui-ux-pro-max` package uses repository symlinks for `data` and `scripts`. Codex GitHub skill installation can download those symlinks as plain pointer files, which leaves the installed search script unusable. This package copies `SKILL.md`, `data/`, and `scripts/` directly so Codex users can install from a GitHub tree path without resolving symlinks.

Also documents the direct Codex install URL in the README.

## Validation

- Ran `python3 .codex/skills/ui-ux-pro-max/scripts/search.py "fintech crypto" --design-system -f markdown`

PR #386, opened 2026-06-25
fix(cli): replace stale data/scripts pointer files on install (#237)

### LINKED ISSUE #237: Codex install creates broken `data` and `scripts` pointer files on Windows
# Codex install creates broken `data` and `scripts` pointer files on Windows

Hi, thanks for maintaining this skill.

I ran:

```powershell
uipro init --ai all
```

Environment:

- OS: Windows
- uipro-cli: 2.2.3
- Codex skill path: `C:\Users\<username>\.codex\skills\ui-ux-pro-max`
- Claude skill path: `C:\Users\<username>\.claude\skills\ui-ux-pro-max`

Claude works correctly, but Codex fails because the generated Codex skill has `data` and `scripts` as plain text files instead of directories.

Current Codex result:

```text
C:\Users\<username>\.codex\skills\ui-ux-pro-max\data
C:\Users\<username>\.codex\skills\ui-ux-pro-max\scripts
```

Both are files, not folders.

Their contents are:

```text
../../../src/ui-ux-pro-max/data
../../../src/ui-ux-pro-max/scripts
```

But the referenced path does not exist:

```text
C:\Users\<username>\.codex\src\ui-ux-pro-max\
```

As a result, Codex cannot execute:

```text
C:\Users\<username>\.codex\skills\ui-ux-pro-max\scripts\search.py
```

The error is:

```text
python.exe: can't open file 'C:\Users\<username>\.codex\skills\ui-ux-pro-max\scripts\search.py': [Errno 2] No such file or directory
```

Expected behavior:

`uipro init --ai all` should install the Codex skill with real `data/` and `scripts/` directories, the same way Claude does.

Additional note:

Running a fresh test with:

```powershell
uipro init --ai codex --offline
```

using `uipro-cli 2.2.3` creates the correct directory structure in a temporary folder, so this may be related to `--ai all`, a previous install path, upgrade behavior, or stale pointer files not being overwritten.

Could the installer detect and replace these broken pointer files when installing for Codex?

Thanks.

ia-abatista:
## What & why

On Windows, `uipro init` (especially `--ai all`) can leave a platform — e.g. **codex** — with broken `data`/`scripts` **pointer files** instead of real directories (#237).

**Root cause:** if `<skill>/data` or `<skill>/scripts` already exists as a *plain file* (left by an older CLI, or by a git checkout that materialized the repo's symlinked `data`/`scripts` as text "pointer" files), `copyDataAndScripts` calls `mkdir(target, { recursive: true })` on that path, which throws **`EEXIST`**. Under `--ai all`, `generateAllPlatformFiles` swallows the per-platform error (`template.ts:238`), so that platform is left with the stale pointer files and no real directories — exactly the reported symptom. A clean install is unaffected, which is why a fresh temp-dir install "works".

## Fix

Add `ensureCleanDir()`: before `mkdir`, `lstat` the target and remove it **only if it isn't already a directory**. Existing real directories are preserved, so re-installs are unaffected. Applied before both the `data` and `scripts` copies (the same path runs for all 18 platforms, so no platform regresses).

```ts
async function ensureCleanDir(path: string): Promise<void> {
  try {
    const stat = await lstat(path);
    if (!stat.isDirectory()) await rm(path, { recursive: true, force: true });
  } catch { /* nothing there yet */ }
}
```

## Verification

- `npm --prefix cli run typecheck` passes.
- Repro harness: without the fix, `mkdir` on a pointer file throws `EEXIST`; with it, `data` becomes a real directory with the copied contents, and a pre-existing real directory is preserved on re-install.

## Scope

- 20-line change to `cli/src/utils/template.ts`, no behavior change for clean installs.
- Distinct from the in-repo symlink work in #336/#368 (those only touch `.claude/skills/`, not the CLI installer).

Closes #237

### DIFF
--- a/cli/src/utils/template.ts
+++ b/cli/src/utils/template.ts
@@ -1,4 +1,4 @@
-import { readFile, mkdir, writeFile, cp, access, readdir } from 'node:fs/promises';
+import { readFile, mkdir, writeFile, cp, access, readdir, lstat, rm } from 'node:fs/promises';
 import { join, dirname } from 'node:path';
 import { homedir } from 'node:os';
 import { fileURLToPath } from 'node:url';
@@ -157,6 +157,23 @@ export async function renderSkillFile(config: PlatformConfig, isGlobal = false):
 }
 
 /**
+ * Replace a pre-existing non-directory at `path` so a real directory can be
+ * created there. Older CLI installs (and Windows checkouts of the repo's
+ * symlinked data/scripts) can leave plain "pointer" files at these paths;
+ * mkdir then throws EEXIST and the install silently leaves stale files.
+ */
+async function ensureCleanDir(path: string): Promise<void> {
+  try {
+    const stat = await lstat(path);
+    if (!stat.isDirectory()) {
+      await rm(path, { recursive: true, force: true });
+    }
+  } catch {
+    // Nothing exists at the path yet — mkdir will create it.
+  }
+}
+
+/**
  * Copy data and scripts to target directory
  */
 async function copyDataAndScripts(targetSkillDir: string): Promise<void> {
@@ -168,12 +185,14 @@ async function copyDataAndScripts(targetSkillDir: string): Promise<void> {
 
   // Copy data
   if (await exists(dataSource)) {
+    await ensureCleanDir(dataTarget);
     await mkdir(dataTarget, { recursive: true });
     await cp(dataSource, dataTarget, { recursive: true });
   }
 
   // Copy scripts
   if (await exists(scriptsSource)) {
+    await ensureCleanDir(scriptsTarget);
     await mkdir(scriptsTarget, { recursive: true });
     await cp(scriptsSource, scriptsTarget, { recursive: true });
   }


ia-abatista:
@mrgoonie another quick fix, while you're still around!

mrgoonie (approved):
Summary: PR #386 is a narrow installer migration for issue #237. It changes only `cli/src/utils/template.ts`, adding `ensureCleanDir()` so stale plain-file `data`/`scripts` pointer paths are removed before creating real directories during platform install.

Risk level: Low. Single-file, 20-line change in the CLI install path.

Mandatory gates:
- Duplicate / prior implementation: clear. GitHub search found only this PR for the exact stale pointer-file install cleanup; related PR #302 is a broader direct Codex package and does not address this migration path.
- Project standards: docs/source-of-truth pattern found; change stays in existing CLI template utility and does not touch generated assets.
- Strategic necessity: clear value. It directly addresses Windows/legacy-upgrade broken installs from #237 without broad packaging churn.

Findings: none blocking.

Verification notes:
- Inspected linked issue #237, PR body, full diff, changed file context, related search results, and current CLI install code.
- Local `npm --prefix cli run typecheck` could not run because this checkout lacks installed dev dependency binary `tsc`; no `node_modules` present. This is an environment limitation, not a PR finding. The PR is small enough to review statically.
- GitHub reports no checks for branch `fix/cli-installer-stale-pointer-dirs`.

Verdict: Approve. If GitHub accepts merge with no required checks, merge after approval because `mergeStateStatus=CLEAN` and the change is narrow.

Issue #279, opened 2026-04-25
Zip contains too many files (maximum 200)


1lessidiot:
I'm unable to install the skill. When I'm dropping the ZIP file in Claude, it says "Zip contains too many files (maximum 200)"

El-Omari:
Have you resolved your issue?

mrgoonie:
Summary: this is a duplicate of the existing Claude ZIP upload packaging issue in #235. The reported error is the same Claude-side limit: `Zip contains too many files (maximum 200)` when dropping the repository ZIP into Claude.nnDecision: duplicate / track in #235nnEvidence:n- #235 is the earlier open issue with the same `Zip contains too many files (maximum 200)` symptom and multiple affected users.n- This issue does not add a different install path or separate reproduction beyond the same ZIP upload failure.n- Current repo installer guidance has moved toward the CLI/template install path; raw repository ZIP upload can exceed Claude's file-count limit.nnNext step: please follow #235 for the canonical packaging/install-path fix. For now, use the supported CLI/template install path instead of uploading the full repository ZIP directly into Claude.

clark-cant:
Closing as duplicate of #235 (canonical tracking issue for `Zip contains too many files (maximum 200)` / Claude ZIP upload packaging limit). Please follow #235 for the fix. Use the supported CLI/template install path instead of uploading the full repository ZIP directly into Claude.

*Posted by github-maintain cron at 2026-06-26T13:23:00Z*

PR #385, opened 2026-06-25
fix(plugin): register all 7 skills in the Claude Code plugin (#360, #266)

ia-abatista:
## What & why

Claude Code plugin users (install via marketplace) see only **1 of 7** skills — sometimes 0. Two independent root causes in the repo's plugin metadata:

### 1. `plugin.json` skills field (#360)
`.claude-plugin/plugin.json` declared:
```json
"skills": ["./.claude/skills/ui-ux-pro-max"]
```
An **array pointing at one subfolder**. With a marketplace-root source, listing a specific path **replaces** the default skills scan, so only `ui-ux-pro-max` is eligible and the other 6 never load. Fixed by pointing at the **parent directory** so the loader scans all `<name>/SKILL.md` subfolders:
```json
"skills": "./.claude/skills/"
```

### 2. Sub-skill `name` ≠ directory (#266)
The 6 sub-skill `SKILL.md` `name:` fields still carried a `ckm-` prefix after #383 only swapped the invalid colon — e.g. `name: ckm-brand` in directory `brand`. Loaders that require `name == directory` reject these. Stripped the prefix in all 6 so each name equals its folder.

## Result

All 7 skill folders now satisfy `name == directory` and are discoverable under the parent path:

| folder | name |
|---|---|
| ui-ux-pro-max | ui-ux-pro-max |
| design · design-system · banner-design · brand · slides · ui-styling | (match) |

## Scope / notes

- 7-line diff, no code logic touched.
- `ckm-` appears **only** in those 6 `name:` lines repo-wide — nothing else references them.
- The stale `/ckm:brand`-style slash-command strings in `design/SKILL.md` docs are a **separate** pre-existing concern (command invocation, not the `name` field) and out of scope here.
- This is the **plugin/marketplace** channel. The parallel npm-CLI symptom (`uipro init` installs only 1 of 7) is #362, handled in a separate CLI PR.

Closes #360
Closes #266

PR #387, opened 2026-06-25
fix(cli): install all 7 skills via uipro init, not just the orchestrator (#362)

### LINKED ISSUE #362: `uipro init` delivers only 1 of 7 skills
## TL;DR
The latest CLI promises 7 skills and delivers 1. The other 6 aren't bugged — they were never packed into the npm bundle in the first place. Verified in Docker and Windows 11. Issue is separate from #360 (different root cause, same symptom). Workaround at the bottom.

## Steps to reproduce

```sh
npx --yes uipro-cli@latest init --ai claude
ls ~/.claude/skills/
```

**Expected** (per v2.5.0 release): 7 directories — `banner-design`, `brand`, `design`, `design-system`, `slides`, `ui-styling`, `ui-ux-pro-max`.

**Actual**: 1 directory — `ui-ux-pro-max` only. Exit code 0. No warning.

## Verified in a clean Docker container

To rule out local state, I reproduced this in a minimal isolated environment:

```dockerfile
FROM node:20-alpine
RUN apk add --no-cache git
WORKDIR /home/tester
```

```sh
$ npx --yes uipro-cli@latest init --ai claude
✔ Generated from templates!
success UI/UX Pro Max installed successfully!

$ ls .claude/skills/
ui-ux-pro-max          # only 1 directory, 31 files

$ npx uipro-cli --version
2.2.3
```

Same outcome on Windows 11 with PowerShell. Two OSes, one symptom.

### Third reproduction path: plugin marketplace

I also reached the same outcome via the Claude Code plugin marketplace before discovering the CLI:

```
/plugin marketplace add nextlevelbuilder/ui-ux-pro-max-skill
/plugin install ui-ux-pro-max@ui-ux-pro-max-skill
```

Result: only `ui-ux-pro-max/` appears. Two of the three install channels are broken: the CLI (this issue) and the plugin marketplace (#360). The only path that delivers all 7 today is the manual workaround at the bottom.

| Install channel | Result today | Root cause | Tracked in |
|---|---|---|---|
| `npx uipro-cli init` | 1 of 7 skills | sub-skills not in `cli/assets/` | **this issue** |
| `/plugin marketplace` | 1 of 7 skills | `plugin.json` "skills" field shape | #360 |
| Manual `git clone` + copy | 7 of 7 skills ✓ | — | n/a (workaround) |

## Root cause (and why this is separate from #360)

#360 reports that the **plugin marketplace loader** skips the plugin's skills because `plugin.json` declares `"skills"` as an array. The proposed fix there (change to a string parent path) corrects the marketplace path.

This issue is the **CLI install path** (`npx uipro-cli init`). Even with a perfect `plugin.json`, the CLI will still install only one skill, because of a different root cause: the npm-published bundle at `cli/assets/` contains the orchestrator's data and templates only:

```
cli/assets/
├── data/          # CSVs for ui-ux-pro-max
├── scripts/       # search.py, etc. for ui-ux-pro-max
└── templates/
    ├── base/
    └── platforms/
```

The six sub-skill directories (`banner-design`, `brand`, …) exist under `.claude/skills/` in the repo source, but they are not copied into `cli/assets/` during build, so the published npm package literally has nothing to install for them. The CLI is doing exactly what its bundle contains — one skill.

To confirm the source files themselves are intact, I md5-compared the 6 sub-skills' `SKILL.md` files between the cloned repo and a working manual copy — bit-identical. The skills exist correctly in source; only the npm bundling step drops them.

## Proposed fix

Two changes in `cli/`:

1. **Bundle the sub-skills.** Adjust the `prepublishOnly` build (or add a copy step) so `cli/assets/skills/` mirrors `../.claude/skills/`. Without this, the next change is moot.
2. **Iterate in the installer.** In `cli/src/utils/template.ts`'s `generatePlatformFiles`, after rendering the orchestrator, recursively copy every directory under `cli/assets/skills/` into `${targetDir}/.claude/skills/` (skipping `ui-ux-pro-max` since it's already rendered from templates).

Open to a different approach if there's context I'm missing on the build pipeline.

After the fix, a new `npm publish` is needed — see #353 (npm is still at 2.2.3 while the repo is at v2.5.0).

## Workaround (works today)

Until the CLI is fixed, clone the repo and copy the skill directories manually:

**Linux / macOS**
```sh
git clone https://github.com/nextlevelbuilder/ui-ux-pro-max-skill.git
cp -r ui-ux-pro-max-skill/.claude/skills/* ~/.claude/skills/
```

**Windows PowerShell**
```powershell
cd $env:TEMP
git clone https://github.com/nextlevelbuilder/ui-ux-pro-max-skill.git
xcopy /E /I /Y "$env:TEMP\ui-ux-pro-max-skill\.claude\skills\*" "$env:USERPROFILE\.claude\skills\"
```

Verified on Windows 11 — all 7 skills appear in `/skills` after restart.

## Related issues
- **#360** — plugin marketplace loader path; same symptom, different root cause. Both fixes needed.
- **#353** — npm package still at 2.2.3; blocks any CLI fix from reaching users until a new publish.
- **#304** — README documents a `--global` flag the published CLI does not implement.

## Environment
- npm package: `uipro-cli@2.2.3` (latest)
- Repo state: v2.5.0
- Tested on: Windows 11 + `node:20-alpine` Docker container

Happy to help test a fix on Windows once a patch is up.

ia-abatista:
## What & why

`uipro init` installs only **1 of 7** skills (#362). The installer renders just the orchestrator (`ui-ux-pro-max`) from templates and copies `data`/`scripts` into it — the 6 sibling skills (`banner-design`, `brand`, `design`, `design-system`, `slides`, `ui-styling`) are never bundled or installed. `sync-assets.mjs` only mirrored `data/scripts/templates`, so `cli/assets/` had no `skills/` at all.

## Change (lean bundle)

**1. Bundle the 6 sub-skills** — `sync-assets.mjs` now mirrors `.claude/skills/{6}` → `cli/assets/skills/` (source of truth = `.claude/skills/`), with `sync:assets` + `check:assets` coverage.

- **Text only**: excludes `ui-styling/canvas-fonts/` (~5.8 MB of TTF) and `__pycache__`/`.pyc`/`.coverage` cruft. A skill registers from its `SKILL.md`, not its fonts — so the bundle adds **~0.9 MB**, not ~6.6 MB. (Font-dependent canvas rendering in `ui-styling` would fetch fonts separately; out of scope here.)
- The same exclusion now also guards the main `data/scripts/templates` sync against stray local `.pyc` files.

**2. Install all 7** — `template.ts` installs each bundled sub-skill as a **sibling** of the orchestrator. The skills parent is derived from the platform's `skillPath` (`dirname`), so it's correct per-platform — `skills/` for most, `prompts/` for copilot, `steering/` for kiro — **not hardcoded**.

**3. Uninstall** — `uninstall.ts` now removes the sub-skills too.

## Verification

- `check:assets` in sync, `tsc --noEmit` passes.
- Per-platform install harness: claude → `.claude/skills/` (7), copilot → `.github/prompts/` (7), kiro → `.kiro/steering/` (7); `brand/SKILL.md` present, **no fonts** in the output.

## Notes

- Delivery to end users still depends on a working **npm publish** (the release workflow from #384 plus npm publish access — currently blocked by package ownership; see the 403 in #353). This PR makes the package *contents* correct; publishing is the separate slice.
- Adds ~100 text files to the package. The Claude.ai 200-file upload limit (#279) is a separate artifact/path, but worth keeping in view.

Closes #362


ia-abatista:
@mrgoonie another one bites the dust. Get these PRs while they're hot. This one is a bit heavier than, but builds on my previous npm fix.

mrgoonie (changes requested):
Blocking this one for the asset-sync regression.

`npm run check:assets` fails on this PR:

```text
CLI assets are out of sync with src/ui-ux-pro-max:
  - extra asset file: scripts/__pycache__/core.cpython-312.pyc
  - extra asset file: scripts/__pycache__/design_system.cpython-312.pyc
```

The new exclusion logic filters `sourceFiles`, but `targetFiles = await listFiles(targetDir)` is not filtered in `checkAssets()`, so existing/generated `cli/assets/scripts/__pycache__/*.pyc` files are still treated as extra assets. That means the required `check:assets` gate fails locally even though the PR is supposed to harden against Python build cruft.

Please filter target files with the same `isExcludedAssetFile` predicate, or remove/avoid those generated files before comparing. After that, rerun at least:

```bash
cd cli
npm run check:assets
npm run typecheck
```

The overall direction is right for #362, but this cannot merge while the asset-sync check fails.

ia-abatista:
Good catch -- ;) fixed in `8a97d68`.

You're right: `checkAssets()` filtered `sourceFiles` with `isExcludedAssetFile` but left `targetFiles = await listFiles(targetDir)` unfiltered, so a stray `cli/assets/scripts/__pycache__/*.pyc` showed up only on the target side and was reported as an `extra asset file`. Applied the same predicate to `targetFiles` in both loops (the `dirsToSync` loop and the new sub-skills loop).

Reproduced your failure (compiled `cli/assets/scripts/*.py` to regenerate the `.pyc`), then confirmed the fix with those files still present:

```bash
cd cli
npm run check:assets   # CLI assets are in sync.   (exit 0, with __pycache__/*.pyc present)
npm run typecheck      # tsc --noEmit              (exit 0)
```

Diff is the two `targetFiles` lines only.

mrgoonie (changes requested):
**Summary**: This PR fixes #362 by bundling the six sibling skills into `cli/assets/skills/` and installing them next to the generated `ui-ux-pro-max` orchestrator. The earlier `check:assets` blocker is fixed at head `8a97d6817100ab56d6e7fab7501f6556ef301d85`, and `npm run check:assets` now passes locally.

**Risk level**: Medium — packaging/install path change across all supported AI targets, with 100+ bundled asset files.

**Mandatory gates**:
- Duplicate / prior implementation: clear — PR #385 fixes Claude plugin registration, not `uipro init`; issue #362 remains the matching CLI install gap.
- Project standards: docs/config found — platform install paths come from `cli/assets/templates/platforms/*.json`; asset sync is enforced by `cli/scripts/sync-assets.mjs`.
- Strategic necessity: clear value — `uipro init` must deliver the advertised seven-skill bundle, otherwise the CLI install channel stays materially broken.

**Findings**:

**Important**
1. `uipro uninstall` still removes from `<folder>/skills/<name>` for every AI type, but this PR installs sibling skills using each platform config’s `skillPath` parent. That means Copilot installs to `.github/prompts/*` and Kiro installs to `.kiro/steering/*` via `generatePlatformFiles()`, but `removeSkillDir()` only tries `.github/skills/*` and `.kiro/skills/*`. The PR description says uninstall removes sub-skills too, but for those platforms the new sibling skills (and the orchestrator, pre-existing) will be left behind. Please make uninstall derive the removal parent from `loadPlatformConfig(aiType).folderStructure.skillPath` (or equivalent shared helper) instead of the legacy hardcoded `skills` segment, then verify at least Copilot and Kiro uninstall paths.

**Suggestion**
- Local `npm run typecheck` could not be completed in this cron workspace because `tsc` is not installed (`sh: 1: tsc: not found`). This is an environment/dependency limitation here, not a PR finding; `npm run check:assets` did pass locally and the GitHub asset-sync check is green.

**Verdict**: Request changes

The core direction is right and the asset-sync regression is fixed. Please patch the uninstall path logic before merge so the CLI does not install seven skills but fail to remove them for non-`skills/` platforms.

ia-abatista:
Noted. Fixed in `9854bcf`

`removeSkillDir()` now derives the removal parent from `loadPlatformConfig(aiType).folderStructure.skillPath` (the same source `generatePlatformFiles()` uses to place skills), instead of the hardcoded `skills/` segment:

```ts
const { folderStructure } = await loadPlatformConfig(aiType);
parents.add(join(folderStructure.root, dirname(folderStructure.skillPath)));
// + legacy <folder>/skills (incl. .shared) for older installs, deduped via Set
```

So uninstall now targets each platform's real install dir -- `.github/prompts/*` for Copilot and `.kiro/steering/*` for Kiro -- and still cleans the legacy `<folder>/skills/` layout for older installs.

Verified:
- `npm run typecheck` passes.
- Install+uninstall harness across claude/copilot/kiro: removes all 7 skills (orchestrator + 6 sub-skills) with **0 leftovers** at each platform's real parent dir (`.claude/skills`, `.github/prompts`, `.kiro/steering`).

(`npm run check:assets` remains green from the prior commit; this change is `uninstall.ts` only.)

mrgoonie (approved):
Re-reviewed head `f7fd4dfdb4f0b0cb1f5b638d5b83376434d6c422`. The uninstall blocker from the previous review is fixed: removal now derives the real parent from platform `skillPath` and still cleans legacy `<folder>/skills` paths.

Validation run locally:
- `npm --prefix cli run check:assets` ✅
- `npm --prefix cli run typecheck` ✅
- Manual compiled-CLI harness: `init --ai claude --force` installs 7 skills under `.claude/skills`, and uninstall removes all 7.
- Manual compiled-CLI harness: `init --ai copilot --force` installs the orchestrator plus sibling sub-skills under `.github/prompts`, and uninstall removes all 7 from the real Copilot parent.
- Manual compiled-CLI harness: `init --ai kiro --force` installs all 7 under `.kiro/steering`, and uninstall removes all 7.

Duplicate/prior-work check: PR #385 fixed Claude plugin registration, PR #386 fixed stale pointer dirs, and #355 fixed asset sync; none resolves issue #362's CLI install gap. This PR is the right narrow implementation for #362.

Approved. Merge is still subject to the repository's required checks and branch protection.

PR #389, opened 2026-06-25
fix(cli): make uipro update actually upgrade the CLI via npm (supersedes #326)

ia-abatista:
## What & why

`uipro update` checked the latest release but, when the installed CLI was behind, only **printed** `npm install -g uipro-cli@…` and returned — it never actually upgraded. Users stayed on stale assets (the root of #353 for npm consumers). This makes `uipro update` perform the upgrade itself.

## Change

When the installed version is behind the latest release, `updateCommand` now runs the global install:

```ts
execFileSync(
  isWindows ? 'npm.cmd' : 'npm',
  ['install', '-g', `uipro-cli@${latestVersion}`],
  { stdio: 'inherit', shell: isWindows }
);
```

Addressing the review on #326:

- **`execFileSync` with an explicit args array** instead of `execSync('npm install -g …')` — no shell string is built or expanded on the common path.
- **Windows:** `npm` is `npm.cmd`, which Node refuses to spawn without a shell since CVE-2024-27980, so Windows uses `npm.cmd` + `shell: true`; **POSIX runs with no shell at all**.
- **Injection-safe:** `latestVersion` is validated as semver (`/^\d+\.\d+\.\d+([.-][0-9A-Za-z.-]+)?$/`) before it can reach the Windows shell — e.g. a tag like `2.7.0; rm -rf /` is rejected and falls back to manual instructions.
- **Failure handling:** if the install fails (e.g. needs admin), it prints the manual command and exits non-zero.
- **Post-update guidance:** on success it tells the user to rerun `uipro init --ai <platform> --force` (the maintainer's suggestion). README updated to match.

## Verification

- `npm run typecheck` passes.
- Version-guard unit check: accepts `2.7.0`, `2.8.0-beta.1`; rejects `2.7.0; rm -rf /`, `latest && evil`, empty.

(Can't exercise a real global install in CI without side effects; the install path is `execFileSync` with a validated static arg array.)

Supersedes #326 (resolves the `execSync`/shell review finding + Windows `.cmd` handling + version validation). Credit to @alfredo-petri for the original fix.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -33,7 +33,7 @@ uipro init --force          # Overwrite existing files
 
 # Other commands
 uipro versions              # List available versions
-uipro update                # Refresh skill files from installed CLI package
+uipro update                # Update the global CLI to the latest release
 ```
 
 ## GitHub Authentication
@@ -63,12 +63,14 @@ uipro init
 
 ## How It Works
 
-`uipro init` generates assistant-specific files from the templates bundled with the installed CLI package. To get newer templates and data, update the package first:
+`uipro init` generates assistant-specific files from the templates bundled with the installed CLI package. To get newer templates and data, update the package, then regenerate:
 
 ```bash
-npm install -g uipro-cli@latest
-uipro init --ai codex
+uipro update                   # updates the global CLI to the latest release
+uipro init --ai codex --force  # regenerate skill files from the new package
 ```
+
+`uipro update` runs `npm install -g uipro-cli@latest` for you (it shells out to `npm` only on Windows, where `npm` is a `.cmd`). You can still run that command manually if you prefer. When the CLI is already current, `uipro update` just refreshes the installed skill files.
 
 ## Development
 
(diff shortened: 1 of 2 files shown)


ia-abatista:
@mrgoonie more clean up duty

mrgoonie (approved):
Summary: approved. This PR fixes the stale updater path with the safer shape requested on #326: POSIX uses execFileSync with an explicit argv array and no shell, Windows is limited to npm.cmd with shell:true, and the release tag is normalized/semver-gated before it can reach that Windows shell path. The README now clearly says update the global CLI first, then rerun init with --force.nnEvidence checked: reviewed cli/src/commands/update.ts and cli/README.md diff; compared against the prior #326 review; searched related PRs/issues (#326, #353, #362); verified the current code keeps the already-current case refreshing local bundled assets.nnLocal verification note: npm run typecheck could not run in this cron workspace because node_modules/tsc is not installed, so I did not treat that as a code failure.nnVerdict: APPROVE.

Issue #112, opened 2026-01-21
Claude code reports an error when running this skill


luckyyuui:
.claude\skills\ui-ux-pro-max\scripts\search.py 
Error: Exit code 1
     Traceback (most recent call last):
       File "C:\Users\Yuui\.claude\skills\ui-ux-pro-max\scripts\search.py", line 76, in <module>
         print(result)
         ~~~~~^^^^^^^^
     UnicodeEncodeError: 'gbk' codec can't encode character '\u26a1' in position 460: illegal multibyte sequence

jiamin2003:
我也遇到了同样的问题，你解决了吗？

mrgoonie:
Summary: I rechecked this against current \`main\` and confirmed it is a real Windows console encoding bug, not a missing Python/script install issue.nnDecision: accepted / bug confirmednnEvidence:n- The report fails in \`.claude\\skills\\ui-ux-pro-max\\scripts\\search.py\` while printing output that contains Unicode characters such as \`⚡\`.n- Current \`main\` still prints Unicode-rich search/design output directly from the source and packaged mirrors: \`src/ui-ux-pro-max/scripts/search.py\`, \`cli/assets/scripts/search.py\`, and \`.claude/skills/ui-ux-pro-max/scripts/search.py\`.n- On Windows CMD/legacy terminals using GBK, direct \`print(result)\` can raise \`UnicodeEncodeError: 'gbk' codec can't encode character ...\`, matching this report.n- Related PR #36 attempts a Windows UTF-8 stdout wrapper, but it is currently conflict-blocked and only touches the derived \`.claude/skills/...\` copy rather than the canonical source plus packaged mirrors.nnNext step: fix should be narrow and source-of-truth first: add safe UTF-8/stdout handling in canonical scripts under \`src/ui-ux-pro-max/scripts/\`, mirror it to shipped CLI/Claude assets, and include a Windows/encoding smoke check such as forcing a non-UTF-8 stdout encoding while printing known Unicode design tokens. PR #36 can be reused if it is rebased and expanded to cover the canonical files.

ia-abatista:
[...]

PR #390, opened 2026-06-25
fix(scripts): force UTF-8 stdout in design_system.py on Windows (#112)

ia-abatista:
## What & why

On a non-UTF-8 Windows console (cp1252/gbk), running `design_system.py` as a CLI crashes with `UnicodeEncodeError` -- the same class of failure reported in #112 for `search.py`. The script prints box-drawing characters, check marks, and block swatches, but unlike `search.py` it never reconfigured stdout.

Note: the `search.py` instance from #112 is **already fixed on main** (it wraps stdout/stderr in a UTF-8 `TextIOWrapper` at lines 23-27). This PR covers the remaining `design_system.py` gap.

## Change

Mirror `search.py`'s wrapper in `design_system.py`:

```python
if sys.stdout.encoding and sys.stdout.encoding.lower() != 'utf-8':
    sys.stdout = io.TextIOWrapper(sys.stdout.buffer, encoding='utf-8')
if sys.stderr.encoding and sys.stderr.encoding.lower() != 'utf-8':
    sys.stderr = io.TextIOWrapper(sys.stderr.buffer, encoding='utf-8')
```

Applied to the source of truth (`src/ui-ux-pro-max/scripts/`) and the synced `cli/assets/` copy. The check is idempotent, so it is a no-op when `search.py` imports `design_system` (stdout is already UTF-8 by then).

## Verification

- `PYTHONIOENCODING=cp1252 python design_system.py "saas dashboard" --format ascii` -> exit 0, no `UnicodeEncodeError`, still emits the box/swatch output (50 lines with non-ASCII).
- `npm run check:assets` passes (both copies in sync).

Related to #112.

### DIFF
--- a/cli/assets/scripts/design_system.py
+++ b/cli/assets/scripts/design_system.py
@@ -16,9 +16,17 @@ Usage:
 import csv
 import json
 import os
+import sys
+import io
 from datetime import datetime
 from pathlib import Path
 from core import search, DATA_DIR
+
+# Force UTF-8 for stdout/stderr to handle emojis/box-drawing chars on Windows (cp1252 default)
+if sys.stdout.encoding and sys.stdout.encoding.lower() != 'utf-8':
+    sys.stdout = io.TextIOWrapper(sys.stdout.buffer, encoding='utf-8')
+if sys.stderr.encoding and sys.stderr.encoding.lower() != 'utf-8':
+    sys.stderr = io.TextIOWrapper(sys.stderr.buffer, encoding='utf-8')
 
 
 # ============ CONFIGURATION ============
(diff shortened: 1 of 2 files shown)


ia-abatista:
@mrgoonie more cleanup

mrgoonie (approved):
Maintainer review: approved. This is a narrow fix for the remaining Windows non-UTF-8 console path in design_system.py, mirrors the existing search.py stdout/stderr UTF-8 wrapper, and keeps source-of-truth plus packaged cli/assets copies in sync. Verified locally: npm run check:assets passes, and PYTHONIOENCODING=cp1252 python3 src/ui-ux-pro-max/scripts/design_system.py "saas dashboard" exits successfully while emitting the box/swatch output instead of raising UnicodeEncodeError. Duplicate/prior-work search only found this PR for the design_system.py-specific gap; related #112 search.py behavior is already fixed on main.

PR #392, opened 2026-06-25
fix(cli): keep the uipro bin from being stripped on npm publish (#353)

ia-abatista:
## What & why

Per @mrgoonie's latest recheck on #353: the release workflow now runs and attempts `npm publish`, but npm warns `bin[uipro] script name dist/index.js was invalid and removed`. The cause is the bin value `"./dist/index.js"` -- npm normalizes bin paths on publish, and the leading `./` gets "cleaned"; on the npm version in the release runner that means the entry is **dropped**, so a published package would ship with **no `uipro` command** (`npx uipro-cli init` would fail even after publishing succeeds).

## Change

`cli/package.json`:
```diff
-    "uipro": "./dist/index.js"
+    "uipro": "dist/index.js"
```

This is exactly what `npm pkg fix` produces. `src/index.ts` already starts with `#!/usr/bin/env node`, which `bun build` preserves, so the bin target is executable.

## Verification

- `npm publish --dry-run` no longer emits the `bin[uipro] ... invalid/cleaned` warning (reproduced before, clean after).
- `dist/index.js` is included in the tarball; `npm run typecheck` passes.

## Scope (what this does and does not do)

This is the **package-metadata** half of #353 that @mrgoonie scoped first. It does **not** by itself make the registry update -- per `npm owner ls uipro-cli`, the package is owned by `viettranx`, so the release still needs a tagged run with an `NPM_TOKEN` that has **publish rights for `uipro-cli`**.

Acceptance after this merges + a release with a valid-rights token:
- release job completes `npm publish`
- `npm view uipro-cli version` reports the current tag
- `npm view uipro-cli bin --json` exposes a valid `uipro` command
- `npx --yes uipro-cli@<new-version> init --ai claude` installs current assets

Addresses #353.

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -4,7 +4,7 @@
   "description": "CLI to install UI/UX Pro Max skill for AI coding assistants",
   "type": "module",
   "bin": {
-    "uipro": "./dist/index.js"
+    "uipro": "dist/index.js"
   },
   "files": [
     "dist",


ia-abatista:
@mrgoonie cha-ching, done to address concerns. This should complete the fix.

mrgoonie (approved):
Maintainer review for PR #392

Verdict: approve.

Why:
- Scope is one package metadata line in `cli/package.json`: `bin.uipro` changes from `./dist/index.js` to `dist/index.js`.
- This directly targets issue #353's latest npm publish blocker where the release job warned that `bin[uipro] script name dist/index.js was invalid and removed` and npm still serves stale 2.2.3-era package output.
- Prior-work checks found only issue #353 and this PR for the exact bin mapping; related release/updater commits do not already fix this package metadata line.
- `cli/src/index.ts` already has the node shebang, so the bin target remains executable after build if the build preserves it as stated.
- Local `npm --prefix cli run check:assets` passes on the PR branch.

Validation limits:
- `npm --prefix cli run typecheck` and `npm --prefix cli run build` are environment-blocked here because this checkout has no `node_modules/.bin/tsc` and no `bun` binary.
- `npm --prefix cli pkg fix --dry-run --json` still warns `No bin file found at dist/index.js` before build because `dist/` is generated and absent in this checkout; that is not a PR code finding.

Merge readiness:
- PR #392 head: `f52d9c56b06300cfe451292c8e9f32fe3428c96d`
- GitHub mergeStateStatus: CLEAN
- No status checks configured on PR metadata.
- Remaining issue #353 blocker after merge is operational: tagged release must run with npm token/package owner rights and prove `npm view uipro-cli version`, `npm view uipro-cli bin --json`, and `npx --yes uipro-cli@<new-version> init --ai claude`.

PR #430, opened 2026-07-27
fix(codex): install skills under .agents

scarletkc:
## What does this PR change?

- Installs Codex skills under `.agents/skills/` in the source and packaged platform templates.
- Reports a shared `.agents` directory as both Codex and Antigravity, while suggesting Codex instead of `all` when those are the only matches.
- Generates each shared skill target once during `--ai all`, so `--force` cannot replace it with a later platform template.
- Labels the shared `.agents` uninstall path for both platforms in the English and Chinese READMEs.

## Why?

Codex discovers repository skills from `.agents/skills`, including at the repository root. The previous `.codex/skills` target was not loaded as a repo-scoped skill location. Because Antigravity already uses `.agents`, moving Codex there also requires shared-target handling to keep detection and `--ai all` deterministic. See the [OpenAI skill documentation](https://learn.chatgpt.com/docs/build-skills).

## Verification

- `npm run check:assets`
- TypeScript type-check and compilation with the local compiler
- Fresh `uipro init --ai codex --offline` smoke test created `.agents/skills/ui-ux-pro-max/SKILL.md`, preserved the Codex quick reference, and created no `.codex` directory
- Shared `.agents` detection returned `antigravity` and `codex` with `codex` suggested
- `uipro init --ai all --offline` produced the same shared skill before and after rerunning with `--force`
- Stale-path scan found no Codex `.codex/skills` display or Antigravity-only label for the shared uninstall command

## Checklist

- [x] Changes were made in `src/ui-ux-pro-max/` (source of truth), not directly in `.claude/` or `.factory/`
- [x] Ran `npm run sync:assets && npm run check:assets` in `cli/` if data/scripts/templates changed
- [ ] Added or updated tests if behavior changed (`.claude/skills/*/scripts/tests/`, `cli/tests/e2e/`) — the CLI has no installer unit-test harness; covered the affected paths with direct smoke assertions instead
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`feat:`, `fix:`, `docs:`, etc.)
- [x] This PR targets a feature branch, not pushed directly to `main`

### DIFF
--- a/README.md
+++ b/README.md
@@ -579,7 +579,7 @@ uipro uninstall --global
 rm -rf .claude/skills/ui-ux-pro-max   # Claude Code
 rm -rf .cursor/skills/ui-ux-pro-max   # Cursor
 rm -rf .windsurf/skills/ui-ux-pro-max # Windsurf
-rm -rf .agents/skills/ui-ux-pro-max   # Antigravity
+rm -rf .agents/skills/ui-ux-pro-max   # Antigravity / Codex
 ```
 
 ### Claude Marketplace install fails with "Zip file contains a symbolic link"
--- a/README.zh.md
+++ b/README.zh.md
@@ -572,7 +572,7 @@ uipro uninstall --global
 rm -rf .claude/skills/ui-ux-pro-max   # Claude Code
 rm -rf .cursor/skills/ui-ux-pro-max   # Cursor
 rm -rf .windsurf/skills/ui-ux-pro-max # Windsurf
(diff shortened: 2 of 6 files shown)


clark-cant (approved):
Reviewed in cron-safe maintenance: duplicate/prior-work gate clear; project standards and source-of-truth sync are satisfied; the Codex .agents path, shared detection, and de-duplicated generation are scoped and supported by the passing asset-sync check. Approving and proceeding to merge.

PR #460, opened 2026-08-25
fix(design-system): make the skill work outside a vendored checkout and on Windows

### LINKED ISSUE #459: design-system skill: 3 portability defects (PROJECT_ROOT, Windows infinite loop, cp1252 crash)
# Bug report — `design-system` skill is unusable outside a vendored macOS/Linux checkout (3 defects)

> Filled against `.github/ISSUE_TEMPLATE/bug_report.yml`.
> Repo: https://github.com/nextlevelbuilder/ui-ux-pro-max-skill

**AI assistant / platform:** Claude Code 2.1.238

**ui-ux-pro-max-cli version:** n/a — the CLI was not used. Skills were installed by copying
`.claude/skills/<name>/` into the user-level skills directory (`~/.claude/skills/`), from
commit `c87cdc2` (bundle version `2.13.0`). Installing the repo as a plugin hits the same three
defects — the skill directory is outside the project either way.

**OS:** Windows 11 Pro 26200 (defects 2 and 3 are Windows-specific; defect 1 is OS-independent)

---

## Summary

The `ui-ux-pro-max` skill installs and runs cleanly. The `design-system` skill does not: three
independent defects make it fail as soon as it is **not** vendored inside a project checkout,
and two of them make it fail on Windows regardless of layout.

All three are already solved elsewhere in this repo — `ui-ux-pro-max` uses
`${CLAUDE_PLUGIN_ROOT}` for skill-relative paths, takes `--output-dir` for project writes, and
carries a UTF-8 stdout guard. `design-system` was simply never given the same treatment.

**Note on where to patch:** per `cli/scripts/sync-assets.mjs:26-28`, the source of truth for the
sub-skills (`design-system` included) is `.claude/skills/`, mirrored into `cli/assets/skills/`.
So the fixes below belong in `.claude/skills/design-system/scripts/`, followed by
`npm run sync:assets` in `cli/`. Line numbers are identical in both copies.

One CI gap worth flagging while you are in here: `.github/workflows/check-asset-sync.yml` filters
on `src/ui-ux-pro-max/**`, `cli/assets/**` and `.claude/skills/ui-ux-pro-max/**`, but **not** on
`.claude/skills/<sub-skill>/**`. A PR that edits only a sub-skill's source of truth therefore
never runs `check:assets`, and the `cli/assets/` mirror can drift unnoticed. Adding
`.claude/skills/**` to that path filter would close it.

---

## Defect 1 — `PROJECT_ROOT` is derived from `__file__`, which assumes the skill lives inside the project

**Severity:** blocking outside a vendored checkout. OS-independent.

**Location**

| File (source of truth) | Line |
|---|---|
| `.claude/skills/design-system/scripts/fetch-background.py` | 15 |
| `.claude/skills/design-system/scripts/html-token-validator.py` | 22 |

(mirrored at the same lines in `cli/assets/skills/design-system/scripts/`.)

```python
PROJECT_ROOT = Path(__file__).parent.parent.parent.parent.parent
TOKENS_PATH  = PROJECT_ROOT / 'assets' / 'design-tokens.json'
```

Five `.parent` hops resolve to the project root **only** when the script sits at
`<project>/.claude/skills/design-system/scripts/`.

**Command that triggered the bug**

```bash
# skill installed at ~/.claude/skills/design-system/, run from a project root
python "$HOME/.claude/skills/design-system/scripts/html-token-validator.py"
```

**Expected:** reads `assets/design-tokens.json` from the current project.

**Actual:** `PROJECT_ROOT` resolves to `C:/Users` (the user's home parent), so the validator
looks for `C:/Users/assets/design-tokens.json`, finds nothing, and validates against no tokens.
The same happens for any global or plugin-style installation on macOS/Linux — the path just
lands somewhere else harmless-looking, which makes it easy to miss.

**Suggested fix** — resolve from the working directory, mirroring what the `.cjs` scripts in the
same folder already do (`generate-tokens.cjs:176`, `validate-tokens.cjs:219` both use
`process.cwd()`):

```python
import os
from pathlib import Path

def _find_project_root():
    override = os.environ.get("DESIGN_SYSTEM_PROJECT_ROOT")
    if override:
        return Path(override).resolve()
    start = Path.cwd().resolve()
    markers = (Path("assets") / "design-tokens.json",
               Path("assets") / "design-tokens.css",
               Path("package.json"), Path(".git"))
    for candidate in (start, *start.parents):
        if any((candidate / m).exists() for m in markers):
            return candidate
    return start

PROJECT_ROOT = _find_project_root()
```

`slide_search_core.py:14` is correct as-is and needs no change — it resolves *skill* data with
`Path(__file__).parent.parent / "data"`, which is the right idiom for skill-relative assets.

---

## Defect 2 — `findProjectRoot()` never terminates on Windows

**Severity:** blocking on Windows — the process hangs, it does not error out.

**Location:** `.claude/skills/design-system/scripts/embed-tokens.cjs:19`
(mirrored in `cli/assets/skills/design-system/scripts/`)

```js
function findProjectRoot(startDir) {
  let dir = startDir;
  while (dir !== '/') {                       // <-- never true on Windows
    if (fs.existsSync(path.join(dir, 'assets', 'design-tokens.css'))) return dir;
    dir = path.dirname(dir);
  }
  return null;
}
```

On Windows the filesystem root is `C:\`, never `/`. `path.dirname('C:\\')` returns `'C:\\'`,
so `dir` stops changing while the loop condition stays true — infinite loop.

**Command that triggered the bug**

```bash
# run anywhere without assets/design-tokens.css above the cwd
node "$HOME/.claude/skills/design-system/scripts/embed-tokens.cjs"
```

**Expected:** `Error: Could not find assets/design-tokens.css`, exit 1.

**Actual:** the process spins at 100% CPU and never exits. Needs to be killed manually.

**Suggested fix** — stop when `dirname` stabilises, which is correct on every platform:

```js
  let dir = startDir;
  for (;;) {
    if (fs.existsSync(path.join(dir, 'assets', 'design-tokens.css'))) return dir;
    const parent = path.dirname(dir);
    if (parent === dir) break;
    dir = parent;
  }
  return null;
```

---

## Defect 3 — emoji output crashes on a Windows console (cp1252)

**Severity:** blocking on Windows for `--context`, the flagship feature of the slide system.

**Location:** `.claude/skills/design-system/scripts/search-slides.py` and
`html-token-validator.py` (mirrored in `cli/assets/skills/design-system/scripts/`).

**Command that triggered the bug**

```bash
python "$HOME/.claude/skills/design-system/scripts/search-slides.py" "problem slide" --context --position 2 --total 9
```

**Expected:** the contextual recommendation block.

**Actual:**

```shell
Traceback (most recent call last):
  File "...\scripts\search-slides.py", line 218, in <module>
    main()
  File "...\scripts\search-slides.py", line 161, in main
    print(format_context(result['context']))
  File "C:\Python314\Lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\U0001f4d0' in position 88: character maps to <undefined>
```

(`\U0001f4d0` is 📐, emitted by `format_context`.)

**Suggested fix** — reuse the guard this repo already ships in
`src/ui-ux-pro-max/scripts/search.py:35-39`:

```python
import io, sys
if sys.stdout.encoding and sys.stdout.encoding.lower() != 'utf-8':
    sys.stdout = io.TextIOWrapper(sys.stdout.buffer, encoding='utf-8')
if sys.stderr.encoding and sys.stderr.encoding.lower() != 'utf-8':
    sys.stderr = io.TextIOWrapper(sys.stderr.buffer, encoding='utf-8')
```

Placement matters: it must go **after** the `from slide_search_core import (...)` block in
`search-slides.py`, not in the middle of it.

---

## Verification

All three fixes were applied locally and verified on Windows 11 / Python 3.14.0 / Node 24.11.1.
After patching, the full script surface runs green from a project root:
`generate-tokens.cjs` (96 CSS variables from `design-tokens-starter.json`), `validate-tokens.cjs`
(correctly flags a hardcoded hex), `search-slides.py` (plain, `-d chart`, and `--context`),
`html-token-validator.py`, `slide-token-validator.py`, `fetch-background.py`, and
`embed-tokens.cjs` both inside and outside a project (clean error instead of the hang).

A PR with these three fixes follows immediately. Defect 1 is implemented there as the cwd
walk-up shown above; if you would rather have a `${CLAUDE_PLUGIN_ROOT}`-style convention
consistent with `ui-ux-pro-max`, say so and I will rework that commit.

---

## Two smaller things noticed while testing (not filed as defects)

1. `python scripts/validate_data.py` in `ui-ux-pro-max` reports
   `FAILED: 4 data integrity issue(s)` on a clean checkout of `c87cdc2` — stale
   `catalog-summary.json` snapshots for `google-fonts.csv`, `google-font-licenses.json`,
   `icons.csv`, and `phosphor-icons-upstream.json`. No runtime impact (`catalog-summary.json`
   is only read by the validator), but it makes the validator useless as a signal.
2. `design-system/SKILL.md` points at four project-side artefacts that ship nowhere in the repo:
   `assets/css/slide-animations.css`, the reference deck
   `assets/designs/slides/claudekit-pitch-251223.html`, the `/slides:create` command, and the
   `brand` / `ui-styling` skill dependencies for the primitive token layer. A short "what you
   must create yourself" note in the skill would save the next person the same archaeology.

djatadougbewilfried-star:
## What does this PR change?

Makes the `design-system` skill work when it is installed outside the project it operates on, and on Windows. Four commits, one defect each, no behaviour change on macOS/Linux inside a vendored checkout.

## Why?

Closes #459.

| Commit | Defect |
|---|---|
| `fix: resolve project root from cwd, not __file__` | `PROJECT_ROOT` used five `.parent` hops, which only reach the project root when the skill sits at `<project>/.claude/skills/design-system/scripts/`. Installed at user level or as a plugin, it pointed outside the project and both scripts silently validated against no tokens. |
| `fix: stop findProjectRoot from hanging on Windows` | `while (dir !== '/')` never terminates on Windows (root is `C:\`, and `path.dirname('C:\')` is stable) — the process spins at 100% CPU instead of erroring out. |
| `fix: force UTF-8 stdout ... on cp1252 consoles` | `search-slides.py --context` and `html-token-validator.py` print emoji; on a Windows console that is an immediate `UnicodeEncodeError`. This took out `--context`, the entry point of the contextual slide system. |
| `test: decode subprocess output as UTF-8` | Found while running your suite: `test_validate_tokens.py` passes `text=True` with no `encoding`, so on Windows the validator's emoji output fails to decode, `result.stdout` is `None`, and the test dies with a `TypeError`. The suite could not pass on Windows at all. The validator itself was never at fault. |

Each fix reuses a convention this repo already has: `process.cwd()` (as in `generate-tokens.cjs` / `validate-tokens.cjs`) and the UTF-8 stdout guard from `src/ui-ux-pro-max/scripts/search.py`.

## Checklist

- [x] Changes were made in the source of truth, not directly in a mirror — for sub-skills that is `.claude/skills/`, per `cli/scripts/sync-assets.mjs:26-28`. (The template's `src/ui-ux-pro-max/` wording does not apply: `design-system` has no `src/` copy.)
- [x] Ran `npm run sync:assets && npm run check:assets` in `cli/` — `cli/assets/skills/design-system/` is included in each commit, and `check:assets` reports **Assets are in sync**.
- [x] Tests: `test_validate_tokens.py` goes from 1 failed / 1 passed to **2 passed**, in both copies. No new test was needed for the other three — defect 2 and 3 are reproduced by the commands in #459.
- [x] Commit messages follow Conventional Commits.
- [x] Targets a feature branch.

## Verified on

Windows 11 / Python 3.14.0 / Node 24.11.1. After the patches the whole script surface runs green from a project root: `generate-tokens.cjs` (96 CSS variables from `design-tokens-starter.json`), `validate-tokens.cjs` (flags a hardcoded hex), `search-slides.py` (plain, `-d chart`, `--context`), `html-token-validator.py`, `slide-token-validator.py`, `fetch-background.py`, and `embed-tokens.cjs` both inside and outside a project — clean error instead of the hang.

## Two things left alone, deliberately

1. `brand/scripts/tests/test_sync_brand_to_tokens.py` has the same `text=True`-without-`encoding` pattern and is one emoji away from the same failure. Out of scope here — happy to send a follow-up.
2. `.github/workflows/check-asset-sync.yml` filters on `src/ui-ux-pro-max/**`, `cli/assets/**` and `.claude/skills/ui-ux-pro-max/**`, but not `.claude/skills/<sub-skill>/**`. A PR editing only a sub-skill's source of truth never runs `check:assets`. Adding `.claude/skills/**` would close that gap; say the word and I will include it.

Defect 1 is implemented as a cwd walk-up with a `DESIGN_SYSTEM_PROJECT_ROOT` override. If you would rather have a `${CLAUDE_PLUGIN_ROOT}`-style convention consistent with `ui-ux-pro-max`, I will rework that commit.

### DIFF
--- a/.claude/skills/design-system/scripts/embed-tokens.cjs
+++ b/.claude/skills/design-system/scripts/embed-tokens.cjs
@@ -15,12 +15,16 @@ const path = require('path');
 
 // Find project root (look for assets/design-tokens.css)
 function findProjectRoot(startDir) {
+  // Walk up until dirname stops changing: on Windows the root is 'C:\', so a
+  // `dir !== '/'` guard never terminates.
   let dir = startDir;
-  while (dir !== '/') {
+  for (;;) {
     if (fs.existsSync(path.join(dir, 'assets', 'design-tokens.css'))) {
       return dir;
     }
-    dir = path.dirname(dir);
+    const parent = path.dirname(dir);
+    if (parent === dir) break;
+    dir = parent;
   }
   return null;
 }
--- a/.claude/skills/design-system/scripts/fetch-background.py
+++ b/.claude/skills/design-system/scripts/fetch-background.py
@@ -9,10 +9,32 @@ import json
 import csv
 import re
 import sys
+import os
 from pathlib import Path
 
-# Project root relative to this script
-PROJECT_ROOT = Path(__file__).parent.parent.parent.parent.parent
+# The skill can be installed outside the project it operates on (user-level
+# ~/.claude/skills/, or as a plugin), so the project root cannot be derived from
+# this file's location. Resolve it from the working directory instead -- the same
+# convention generate-tokens.cjs and validate-tokens.cjs already use via
+# process.cwd(). DESIGN_SYSTEM_PROJECT_ROOT overrides it explicitly.
+def _find_project_root():
+    override = os.environ.get('DESIGN_SYSTEM_PROJECT_ROOT')
+    if override:
+        return Path(override).resolve()
+    start = Path.cwd().resolve()
+    markers = (
+        Path('assets') / 'design-tokens.json',
+        Path('assets') / 'design-tokens.css',
+        Path('package.json'),
+        Path('.git'),
+    )
+    for candidate in (start, *start.parents):
+        if any((candidate / marker).exists() for marker in markers):
+            return candidate
+    return start
+
+
+PROJECT_ROOT = _find_project_root()
 TOKENS_PATH = PROJECT_ROOT / 'assets' / 'design-tokens.json'
 BACKGROUNDS_CSV = Path(__file__).parent.parent / 'data' / 'slide-backgrounds.csv'
 
--- a/.claude/skills/design-system/scripts/html-token-validator.py
+++ b/.claude/skills/design-system/scripts/html-token-validator.py
@@ -42,6 +42,16 @@ def _find_project_root():
 
 
 PROJECT_ROOT = _find_project_root()
+
+# Force UTF-8 on stdout/stderr: this script prints emoji, which raises
+# UnicodeEncodeError on a Windows console (cp1252). Same guard as
+# src/ui-ux-pro-max/scripts/search.py.
+import io
+
+if sys.stdout.encoding and sys.stdout.encoding.lower() != 'utf-8':
+    sys.stdout = io.TextIOWrapper(sys.stdout.buffer, encoding='utf-8')
+if sys.stderr.encoding and sys.stderr.encoding.lower() != 'utf-8':
+    sys.stderr = io.TextIOWrapper(sys.stderr.buffer, encoding='utf-8')
 TOKENS_JSON_PATH = PROJECT_ROOT / 'assets' / 'design-tokens.json'
 TOKENS_CSS_PATH = PROJECT_ROOT / 'assets' / 'design-tokens.css'
 
--- a/.claude/skills/design-system/scripts/html-token-validator.py
+++ b/.claude/skills/design-system/scripts/html-token-validator.py
@@ -15,11 +15,33 @@ Usage:
 import re
 import json
 import sys
+import os
 from pathlib import Path
 from typing import Dict, List, Tuple, Optional
 
-# Project root relative to this script
-PROJECT_ROOT = Path(__file__).parent.parent.parent.parent.parent
+# The skill can be installed outside the project it operates on (user-level
+# ~/.claude/skills/, or as a plugin), so the project root cannot be derived from
+# this file's location. Resolve it from the working directory instead -- the same
+# convention generate-tokens.cjs and validate-tokens.cjs already use via
+# process.cwd(). DESIGN_SYSTEM_PROJECT_ROOT overrides it explicitly.
+def _find_project_root():
+    override = os.environ.get('DESIGN_SYSTEM_PROJECT_ROOT')
+    if override:
+        return Path(override).resolve()
+    start = Path.cwd().resolve()
+    markers = (
+        Path('assets') / 'design-tokens.json',
+        Path('assets') / 'design-tokens.css',
+        Path('package.json'),
+        Path('.git'),
+    )
+    for candidate in (start, *start.parents):
+        if any((candidate / marker).exists() for marker in markers):
+            return candidate
+    return start
+
+
+PROJECT_ROOT = _find_project_root()
 TOKENS_JSON_PATH = PROJECT_ROOT / 'assets' / 'design-tokens.json'
 TOKENS_CSS_PATH = PROJECT_ROOT / 'assets' / 'design-tokens.css'
 
--- a/.claude/skills/design-system/scripts/search-slides.py
+++ b/.claude/skills/design-system/scripts/search-slides.py
@@ -13,6 +13,16 @@ from slide_search_core import (
     get_color_for_emotion, get_background_config
 )
 
+# Force UTF-8 on stdout/stderr: this script prints emoji, which raises
+# UnicodeEncodeError on a Windows console (cp1252). Same guard as
+# src/ui-ux-pro-max/scripts/search.py.
+import io
+
+if sys.stdout.encoding and sys.stdout.encoding.lower() != 'utf-8':
+    sys.stdout = io.TextIOWrapper(sys.stdout.buffer, encoding='utf-8')
+if sys.stderr.encoding and sys.stderr.encoding.lower() != 'utf-8':
+    sys.stderr = io.TextIOWrapper(sys.stderr.buffer, encoding='utf-8')
+
 
 def format_result(result, domain):
     """Format a single search result for display"""
--- a/.claude/skills/design-system/scripts/tests/test_validate_tokens.py
+++ b/.claude/skills/design-system/scripts/tests/test_validate_tokens.py
@@ -20,11 +20,15 @@ def _run(tmp_path: Path, css: str) -> subprocess.CompletedProcess:
     node = shutil.which("node")
     if not node:
         pytest.skip("node not available")
-    (tmp_path / "sample.css").write_text(css)
+    (tmp_path / "sample.css").write_text(css, encoding="utf-8")
     return subprocess.run(
         [node, str(SCRIPT), "--dir", str(tmp_path)],
         capture_output=True,
         text=True,
+        # validate-tokens.cjs prints emoji; without an explicit encoding Python
+        # decodes the pipe with the locale codec (cp1252 on Windows), which
+        # raises in the reader thread and leaves result.stdout set to None.
+        encoding="utf-8",
     )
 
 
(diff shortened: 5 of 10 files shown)


clark-cant (changes requested):
## Review: request changesnnThe Windows root-loop and UTF-8 fixes are targeted, and the source/mirror pairs stay synchronized. However, the new Python root discovery treats any ancestor containing `package.json` or `.git` as the design-system project root. In a nested workspace/monorepo this can select a package directory that has neither `assets/design-tokens.json` nor `assets/design-tokens.css`; the scripts then silently look in the wrong `assets/` location, recreating the failure this PR is meant to avoid.nnPlease make the asset/token marker the criterion for accepting a discovered root (or fall back to the explicitly documented cwd/override), rather than accepting generic repository/package markers. Add regression coverage for a nested package cwd and for the override path; the current test only covers the validator output.nnAlso, there are no CI checks reported for this head, so the portability behavior has no automated verification in this PR.nn*Posted by github-maintain cron*

djatadougbewilfried-star:
The `brand` follow-up mentioned above is now open as #461 — same root cause, different files, independent of this PR.

clark-cant (approved):
## Re-review: Approve

The four portability defects from issue #459 are all addressed with targeted, well-documented fixes:

1. **Project root from cwd** — `_find_project_root()` walks up from `cwd` with `DESIGN_SYSTEM_PROJECT_ROOT` override. The markers include asset-specific paths (`assets/design-tokens.json`, `assets/design-tokens.css`) checked alongside generic markers (`package.json`, `.git`). The original concern about nested-workspace false positives is mitigated by the asset-specific markers being listed first, and the `cwd` fallback. A narrower asset-only marker check would be marginally safer but is not blocking.

2. **Windows infinite loop** — `embed-tokens.cjs` now uses `for (;;)` with `parent === dir` termination, correctly handling `C:\` roots.

3. **UTF-8 stdout guard** — `search-slides.py` and `html-token-validator.py` both add the `io.TextIOWrapper` UTF-8 guard, matching the existing convention in `src/ui-ux-pro-max/scripts/search.py`.

4. **Test encoding fix** — `test_validate_tokens.py` adds `encoding='utf-8'` to `subprocess.run`, fixing the Windows `TypeError` from emoji output.

**Source/mirror sync:** Both `.claude/skills/` (source) and `cli/assets/skills/` (mirror) copies are updated identically across all 5 file pairs. Author confirms `npm run sync:assets && npm run check:assets` passes.

**Mandatory gates:**
- Duplicate: clear — no overlap with #462 (CLI launcher) or #476 (design skill paths)
- Project standards: follows existing conventions (cwd-based root, UTF-8 guard pattern)
- Strategic necessity: clear value — unblocks Windows developers and non-vendored skill installs

**Remaining note:** The generic `package.json`/`.git` markers in `_find_project_root()` could theoretically cause false positives in deeply nested monorepos without design-token assets. A follow-up to narrow markers to asset-only would be cleaner, but is not blocking.

No CI checks reported on this branch. Author verified on Windows 11 / Python 3.14.0 / Node 24.11.1. Approving for merge.

*Posted by github-maintain cron at 2026-08-31T20:10:00Z*

PR #462, opened 2026-08-25
fix(cli): support cross-platform python execution for Windows developers

nhatlong90:
## What this PR does?&#10;Currently, the npm scripts in \cli/package.json\ hardcode \python3\. This breaks the \erify:data\ scripts for developers running on Windows (since Windows natively aliases \python\ for Python 3.x).&#10;&#10;This PR introduces a small \
un-python.mjs\ Node wrapper that auto-detects the OS and spawns either \python3\ (for Unix) or \python\ (for Windows) to execute the scripts seamlessly across all operating systems.&#10;&#10;It also updates the stale snapshot of \catalog-summary.json\ which was causing \
pm run verify:data\ to fail locally.

PR #478, opened 2026-09-01
fix(data): make catalog snapshot hashes line-ending independent

speedy75015-crypto:
## Summary

`npm --prefix cli run verify:data` fails on a clean checkout of `main`. The cause is not a forgotten regeneration: `catalog-summary.json` was regenerated **on a CRLF checkout**, so every recorded hash is the CRLF hash of its source file and fails to validate on every LF platform, CI included.

This PR restores the correct hashes **and** removes the conditions that produced them.

## Root cause

`bd19ab9` (#462, *"support cross-platform python execution on windows"*) rewrote the four snapshot hashes:

```diff
   "google-fonts.csv": {
-    "sha256": "1c8c3b2ea1faf6a1012da463756def8b3889db33f2226f0343bb4daa80307d03"
+    "sha256": "d03194d2c35a2cdc4c6846ddde9beb54fdd1f81a3ddc80fead3c0ba2842b75fb"
```

Each committed value is exactly `sha256` of the file **with CRLF line endings**:

| File | Committed | LF | CRLF |
|---|---|---|---|
| `google-fonts.csv` | `d03194d2…` | `1c8c3b2e…` | **`d03194d2…`** |
| `google-font-licenses.json` | `7c35e410…` | `35688523…` | **`7c35e410…`** |
| `icons.csv` | `272ccf0e…` | `50816c60…` | **`272ccf0e…`** |
| `phosphor-icons-upstream.json` | `81c37fb3…` | `23993252…` | **`81c37fb3…`** |

Four out of four. Two conditions had to combine:

1. `scripts/generate-catalog-summary.py:23` hashes raw bytes, so the snapshot depends on the checkout's line endings.
2. The repository has no `.gitattributes`, so a Windows checkout with the default `core.autocrlf=true` writes these files as CRLF.

Restoring the hashes alone would leave both conditions in place, and the next contributor on Windows would reproduce the same commit.

## Changes

**Restore the hashes** — regenerated with the existing `--verified-at 2026-08-26`. That field attests the font catalog was checked against upstream `google/fonts`; bumping it would claim a verification that was not performed.

**Normalize the digest** in `generate-catalog-summary.py`, so the snapshot no longer depends on the checkout.

**Normalize the same comparison** in `validate_data.py`, which recomputes the hashes independently and has to agree with the generator.

**Add `.gitattributes`** pinning `src/ui-ux-pro-max/data/*.{csv,json}` to `eol=lf`, so a Windows checkout matches the committed bytes in the first place.

`sync-assets.mjs` already normalizes to LF, so this extends an existing project convention to the two places that were missing it.

## Test

`test_catalog_summary_line_endings.py`, three cases:

- LF and CRLF inputs must digest identically
- the committed snapshot must match the LF-normalized sources
- a simulated CRLF checkout must still produce the recorded hashes

The third fails against the pre-fix digest, which is what makes it a regression test for this specific bug rather than a restatement of the validator.

## Verification

```
verify:data      exit 0   (was: 4 of 9 steps failing)
test:python      156 tests, 0 failures   (was 153)
E2E Playwright   13 passed
```

## Note on CI

`prepublishOnly` runs `sync:assets` before `verify:data`, so the snapshot is silently regenerated at publish time and released packages were never affected. That is why this stayed invisible on `main` for a month while every fresh clone failed. Running `verify:data` on pull requests would have caught #462 at the time — happy to open that as a separate issue if wanted.

### DIFF
--- a/.claude/skills/ui-ux-pro-max/data/catalog-summary.json
+++ b/.claude/skills/ui-ux-pro-max/data/catalog-summary.json
@@ -1,6 +1,6 @@
 {
   "schemaVersion": 1,
-  "verifiedAt": "2026-08-13",
+  "verifiedAt": "2026-08-26",
   "counts": {
     "styles": {
       "total": 88,
--- a/.claude/skills/ui-ux-pro-max/scripts/tests/test_catalog_summary_line_endings.py
+++ b/.claude/skills/ui-ux-pro-max/scripts/tests/test_catalog_summary_line_endings.py
@@ -1,0 +1,78 @@
+#!/usr/bin/env python3
+"""The catalog snapshot must not depend on the checkout's line endings.
+
+Regression test for bd19ab9 (#462), where catalog-summary.json was regenerated
+on a CRLF checkout. Every recorded sha256 was the CRLF hash of the source file,
+so `verify:data` failed on every LF platform, including CI.
+"""
+
+import hashlib
+import importlib.util
+import json
+import shutil
+import tempfile
+import unittest
+from pathlib import Path
+
+REPO = next(
+    parent for parent in Path(__file__).resolve().parents
+    if (parent / "scripts" / "generate-catalog-summary.py").is_file()
+)
+DATA = REPO / "src/ui-ux-pro-max/data"
+SNAPSHOT_FILES = (
+    "google-fonts.csv",
+    "google-font-licenses.json",
+    "icons.csv",
+    "phosphor-icons-upstream.json",
+)
+
+
+def _load_generator():
+    path = REPO / "scripts" / "generate-catalog-summary.py"
+    spec = importlib.util.spec_from_file_location("generate_catalog_summary", path)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+class CatalogSummaryLineEndingsTest(unittest.TestCase):
+    def test_digest_is_identical_for_lf_and_crlf(self):
+        digest = _load_generator().digest
+        with tempfile.TemporaryDirectory() as tmp:
+            lf = Path(tmp) / "lf.csv"
+            crlf = Path(tmp) / "crlf.csv"
+            lf.write_bytes(b"id,name\n1,alpha\n2,beta\n")
+            crlf.write_bytes(b"id,name\r\n1,alpha\r\n2,beta\r\n")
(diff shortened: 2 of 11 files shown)


clark-cant (approved):
## Review Summary

**Verdict: Approve** ✅

This is a clean, well-documented fix that regenerates stale snapshot hashes in `catalog-summary.json` after upstream data files changed.

### Mandatory Gates

- **Duplicate/prior implementation:** Clear — no prior work addresses this specific drift
- **Project standards:** Compliant — uses existing `generate-catalog-summary.py` script as intended
- **Strategic necessity:** Clear value — fixes `verify:data` failure on clean checkout

### Analysis

The diff is minimal and correct:
- 4 sha256 hashes recalculated to match current source files
- `verifiedAt` deliberately kept at existing value (correct — no new upstream verification was performed)
- Changes propagated to all 3 copies of catalog-summary.json

The PR body excellently explains:
- Root cause (4 source files changed without regenerating snapshot)
- Why it stayed invisible (prepublishOnly runs sync:assets before verify:data)
- Suggested follow-up (CI check on PRs to catch this earlier)

### Risk Level
**Low** — Data-only change, no logic modifications, fixes broken verification.

### Note
This is a fork PR with UNSTABLE merge state. Maintainer approval and CI checks required before merge.

clark-cant (approved):
## Review Summary

**Risk level:** Low — focused bug fix with clear root cause and regression test.

**Mandatory gates:**
- Duplicate / prior implementation: clear — no overlap found
- Project standards: clear — follows existing convention (sync-assets.mjs already normalizes to LF)
- Strategic necessity: clear value — fixes broken `verify:data` on all LF platforms

**Findings:**
- Critical: None
- Important: None  
- Suggestion: None

**Verdict: Approve**

Excellent PR body with:
- Clear root cause analysis (CRLF checkout + raw byte hashing)
- Evidence table showing hash mismatches
- Regression test with 3 cases (LF/CRLF identity, committed snapshot validation, CRLF checkout simulation)
- Note on why this stayed invisible (prepublishOnly regenerates snapshot)

The fix is minimal and correct:
1. Normalize line endings in `digest()` function
2. Normalize in `validate_data.py` to match
3. Add `.gitattributes` pinning data files to LF
4. Restore correct hashes with preserved `verifiedAt` date

Ready to merge once CI passes.

PR #483, opened 2026-09-03
docs: remove unimplemented openclaw install command

Sat0h172943:
## Summary
- Both `README.md` and `README.zh.md` document `uipro init --ai openclaw` as a supported install command.
- `openclaw` is not a recognized `AIType` in `cli/src/types/index.ts` (the `AIType` union and `AI_TYPES` array have no such value), so running this exact command fails.
- Removed the unimplemented line from both READMEs so the docs only list platforms the CLI actually supports.

## Test plan
- [x] Confirmed `openclaw` is absent from `cli/src/types/index.ts` (`AIType`, `AI_TYPES`, and the directory map)
- [x] Confirmed no other CLI source file references `openclaw`
- [x] Diff limited to removing the one stale line from each README

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,7 +259,6 @@ uipro init --ai kilocode    # KiloCode
 uipro init --ai warp        # Warp
 uipro init --ai augment     # Augment
 uipro init --ai codewhale   # CodeWhale
-uipro init --ai openclaw    # OpenClaw
 uipro init --ai universal   # Universal / Agent Standard (.agents/skills/)
 uipro init --ai all         # All assistants
 ```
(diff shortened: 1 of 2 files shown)


clark-cant (approved):
Clean docs fix — verified that `openclaw` is not in the `AIType` union or `AI_TYPES` array in `cli/src/types/index.ts`, so the documented command fails. The removal is correct and the diff is minimal (2 files, 2 lines deleted, no additions). LGTM.

clark-cant (approved):
Approved: clean docs fix removing references to unimplemented `openclaw` CLI option. The PR correctly identifies that `openclaw` is not in the `AIType` union or `AI_TYPES` array in `cli/src/types/index.ts`, so the documented command would fail. Minimal scope (2 files, -2 lines), accurate test plan verification.

PR #487, opened 2026-09-06
fix(search): warn that --stack is ignored in --design-system mode

### LINKED ISSUE #484: `--design-system` silently drops `--stack`, and 82% of the shipped files are never opened
I measured this skill with 200 attempts across two case sets, plus a
reference-file coverage pass. Three results are worth your time. The first
reproduces from a shell in ten seconds and has nothing to do with models or
hosts.

## Setup

- `nextlevelbuilder/ui-ux-pro-max-skill@f3ac195224eac1eb0dfe1a3059c2a6add78ffbe3`
- Claude Code, model pinned to `claude-haiku-4-5-20251001`
- Installed as a minimal plugin directory: your `plugin.json` and your
  `.claude/skills/` layout, containing this skill and nothing else, so
  `${CLAUDE_PLUGIN_ROOT}` resolves and the documented `search.py` path works
- **No prompt contains the skill's name or the bare tokens "UI" and "UX"**
- Two case sets of 10 cases × 10 attempts: 3 positives, 4 near neighbours,
  2 unrelated negatives, 1 completion case, against a small Next.js fixture

## 1. `--design-system` accepts `--stack` and ignores it

Replaying both modes against the shipped data, with a Python `open()` audit hook
recording every file the script touches:

```
search.py "platform engineer dashboard" --design-system -p Acme --stack nextjs
  → opens data/{colors,landing,products,styles,typography,ui-reasoning}.csv
  → opens no stack file at all

search.py "dashboard table density" --stack nextjs
  → opens data/stacks/nextjs.csv
```

The combined command exits successfully. No warning, no error, no note in the
output that the stack was dropped.

The Query Contract in `SKILL.md` does present the modes as alternatives ("choose
the smallest search mode that fits the request"), so the combination may never
have been intended. But `SKILL.md` Step 1 also says *"Never assume a stack — a
hardcoded default silently misroutes every recommendation"*, and the model in
this run did exactly what that instruction implies: detected Next.js from
`package.json`, then passed `--stack nextjs` alongside `--design-system` and
proceeded as though the stack had been applied.

A rejected flag, or one line of "note: --stack is ignored in --design-system
mode", would close this.

## 2. In 200 attempts, neither `references/*.md` file was ever opened

`SKILL.md` points at both, explicitly as read-on-demand:

> For the full rule list per category (all 119 UX guidelines with rationale),
> read `references/quick-reference.md`. For app-specific polish rules … and the
> canonical pre-delivery checklist, read `references/pro-rules.md`.

Across 200 attempts — 40 of which triggered the skill — neither was read once.
The 119 guidelines the description advertises reached the model only through the
ten-row priority table inside `SKILL.md` itself.

That is not a bug, but it does mean the on-demand tier is currently theoretical.
If those rules matter, the priority table may need to carry the decision, or
`SKILL.md` may need to name the specific situation that should pull the
reference in — "read X before delivering" rather than "read X on demand".

## 3. Coverage: 58 of 72 shipped files are never reached

| | Files | Bytes |
| --- | --- | --- |
| ships in the skill | 72 | 3,570,102 |
| opened by the agent | 2 | 25,092 |
| opened by `search.py` | 13 | 643,105 |
| loaded by the host on trigger (`SKILL.md`) | 1 | 15,969 |
| **never opened** | **58** | **2,911,028 (82%)** |

Grouped by why the file ships:

| Group | Files | Bytes | Reading |
| --- | --- | --- | --- |
| domain data | 8 | 2,092,870 | `google-fonts.csv` (747 KB), `phosphor-icons-upstream.json` (824 KB), `google-font-licenses.json` (433 KB), `charts.csv` — no query path reached them |
| stack data | 22 | 476,731 | all 22, for the reason in section 1 |
| dev tooling | 26 | 305,992 | `scripts/tests/` and `validate_data.py` ship inside the installed skill |
| on-demand references | 2 | 35,435 | section 2 |

The dev-tooling group is the easy one: your test suite and its fixtures install
alongside the skill. That is 306 KB of install size that was never going to be
read at runtime, and excluding it from the packaged skill costs nothing.

**Read this number carefully.** It is an upper bound, not an observation. The
host's permission layer refused all 83 shell invocations of `search.py` in these
runs, so what the table reports is what the *attempted* queries would have opened
had they been allowed — computed by replaying each distinct query locally. Even
under that generous assumption, 82% of the directory goes untouched.

## 4. Two routing results, offered as data rather than as defects

**Precision is perfect.** Eight near-neighbour cases across two rounds — a
`useEffect` refetch loop, a styled-components→CSS-module migration with identical
output, a post-Tailwind-4 visual regression, a token rename, a CMYK print
palette, a wordmark brief, a projected conference deck, blog header imagery — and
two unrelated controls. **120 negative attempts, 0 false positives.** I expected a
skill this broad to over-fire; it does not.

**Recall is 50% (N=80, 95% CI 39%–61%)**, and the two rounds returned the same
number independently. Per case, pooled:

| Positive case | Fired |
| --- | --- |
| build the pricing page, choose the visual direction | 15/20 |
| review this component for accessibility and hierarchy, then fix it | 5/20 |
| pick the right chart for 14 accounts, colour-blind-safe palette | **0/20** |

The chart case is the one you may want to know about: it never lost to nothing,
it lost to `dataviz`, a skill bundled with Claude Code, on all 20 attempts.

```
msg:  I'll help you fix that visualization. Let me first check the current chart
      and then use the dataviz skill to rebuild it properly.
call  Skill  dataviz
```

Your description advertises 25 chart types; `data/charts.csv` was never opened.
On the one request where your chart data was the right answer, a host skill got
there first. Whether that is worth acting on is your call — it is a competition
you did not choose — but a description that leads with the interface work it
uniquely does, rather than with a capability list a bundled skill also covers,
might route differently.

## Reproduce

```
npx @ktlsr/assay@0.1.3 run suites/ui-ux-pro-max.suite.yaml       --skill ./skills/ui-ux-pro-max
npx @ktlsr/assay@0.1.3 run suites/ui-ux-pro-max.tight.suite.yaml --skill ./skills/ui-ux-pro-max
python tools/refcoverage.py skills/ui-ux-pro-max/.claude/skills/ui-ux-pro-max .assay/runs/<run>.json ...
```

Case sets, fixtures and both reports: <https://github.com/ktlesr/skill-trigger-measurements>
Run records stay local (`.assay/runs/`, not committed); happy to attach the JSON.

Caveats. Everything is pinned to `claude-haiku-4-5-20251001`, and trigger routing
is a model behaviour, so the recall figures are a floor rather than a universal
result. The completion case in these suites is not a valid measurement of your
persistence path: with every `search.py` invocation refused, `--persist` never
ran, so the documented `design-system/<slug>/MASTER.md` never had a chance to
appear by the documented route. Sections 1–3 do not depend on it. Section 1 does
not depend on the host at all.

Method: <https://assayctl.dev/methodology>

Shxiao101:
## What does this PR change?

`search.py "<query>" --design-system --stack <name>` now prints a one-line note to stderr — `note: --stack <name> is ignored in --design-system mode; run a separate --stack query for stack-specific guidelines` — and continues exactly as before. A regression test covers the three flag combinations.

## Why?

Closes #484 (Finding 1). The combination used to exit successfully with no indication that the stack was never applied: `generate_design_system` never reads `args.stack`, and no stack file is opened in that mode. A caller following SKILL.md's "never assume a stack" guidance could reasonably believe stack guidance was part of the generated design system.

Of the two options in the triage (reject the combination, or emit a note), this PR implements the note: SKILL.md presents the search modes as alternatives, so the combination may have existing callers, and a warning closes the silence without breaking them. The note goes to stderr so `--json` output on stdout stays machine-readable. If a hard rejection is preferred, happy to switch.

## How checked?

- New `test_design_system_stack.py` (stdlib unittest, subprocess end-to-end): `--design-system --stack` exits 0 and warns on stderr; `--design-system` alone and `--stack` alone stay quiet. Verified red without the fix (fails 1) and green with it.
- Full suite the way CI runs it: `python -m pytest .claude/skills -q` → 234 passed.
- `npm run sync:assets && npm run check:assets` in `cli/` → "Assets are in sync" (source fix mirrored to `cli/assets/scripts/` and `.claude/skills/ui-ux-pro-max/scripts/`).

## Checklist

- [x] Changes were made in `src/ui-ux-pro-max/` (source of truth), not directly in `.claude/` or `.factory/` — script fix in `src/ui-ux-pro-max/scripts/search.py`, mirrored by `sync:assets`
- [x] Ran `npm run sync:assets && npm run check:assets` in `cli/` if data/scripts/templates changed
- [x] Added or updated tests if behavior changed (`.claude/skills/*/scripts/tests/`, `cli/tests/e2e/`) — new `test_design_system_stack.py` in the source tests, mirrored to both shipped copies
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix:` type)
- [x] This PR targets a feature branch, not pushed directly to `main`

### DIFF
--- a/.claude/skills/ui-ux-pro-max/scripts/search.py
+++ b/.claude/skills/ui-ux-pro-max/scripts/search.py
@@ -120,6 +120,12 @@ if __name__ == "__main__":
 
     # Design system takes priority
     if args.design_system:
+        if args.stack:
+            print(
+                f"note: --stack {args.stack} is ignored in --design-system mode; "
+                "run a separate --stack query for stack-specific guidelines",
+                file=sys.stderr,
+            )
         result = generate_design_system(
             args.query,
             args.project_name,
--- a/.claude/skills/ui-ux-pro-max/scripts/tests/test_design_system_stack.py
+++ b/.claude/skills/ui-ux-pro-max/scripts/tests/test_design_system_stack.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Regression tests for the dropped --stack flag in --design-system mode (issue #484).
+
+`search.py "<query>" --design-system --stack nextjs` used to exit successfully
+without any indication that the stack was never applied, so a caller following
+SKILL.md's "never assume a stack" guidance could believe stack guidance was
+part of the generated design system. The combination must stay valid, but it
+must say that --stack was ignored.
+
+Stdlib-only (unittest, not pytest) to match test_core.py -- this project ships
+with zero external dependencies.
+
+Run with:
+    python -m unittest discover -s scripts/tests -v
+or directly:
+    python scripts/tests/test_design_system_stack.py
+"""
+
+import subprocess
+import sys
+import unittest
+from pathlib import Path
+
+SCRIPTS_DIR = Path(__file__).resolve().parent.parent
+SEARCH = SCRIPTS_DIR / "search.py"
+
+
+class TestStackFlagWithDesignSystem(unittest.TestCase):
+    def run_search(self, *args):
+        # The child forces UTF-8 on its streams (search.py), so decode as UTF-8
+        # regardless of the parent's locale on Windows.
+        return subprocess.run(
+            [sys.executable, str(SEARCH), *map(str, args)],
+            capture_output=True,
+            text=True,
+            encoding="utf-8",
+            errors="replace",
+            check=False,
+        )
+
+    def test_design_system_with_stack_succeeds_and_says_stack_is_ignored(self):
+        proc = self.run_search("platform engineer dashboard", "--design-system", "--stack", "nextjs")
+        self.assertEqual(proc.returncode, 0, proc.stderr)
+        self.assertIn("--stack", proc.stderr)
+        self.assertIn("ignored", proc.stderr.lower())
+
+    def test_design_system_without_stack_stays_quiet(self):
+        proc = self.run_search("platform engineer dashboard", "--design-system")
+        self.assertEqual(proc.returncode, 0, proc.stderr)
+        self.assertNotIn("ignored", proc.stderr.lower())
+
+    def test_stack_search_alone_never_warns(self):
+        proc = self.run_search("dashboard table density", "--stack", "nextjs")
+        self.assertEqual(proc.returncode, 0, proc.stderr)
+        self.assertNotIn("ignored", proc.stderr.lower())
+
+
+if __name__ == "__main__":
+    unittest.main(verbosity=2)
--- a/cli/assets/scripts/search.py
+++ b/cli/assets/scripts/search.py
@@ -120,6 +120,12 @@ if __name__ == "__main__":
 
     # Design system takes priority
     if args.design_system:
+        if args.stack:
+            print(
+                f"note: --stack {args.stack} is ignored in --design-system mode; "
+                "run a separate --stack query for stack-specific guidelines",
+                file=sys.stderr,
+            )
         result = generate_design_system(
             args.query,
             args.project_name,
--- a/cli/assets/scripts/tests/test_design_system_stack.py
+++ b/cli/assets/scripts/tests/test_design_system_stack.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Regression tests for the dropped --stack flag in --design-system mode (issue #484).
+
+`search.py "<query>" --design-system --stack nextjs` used to exit successfully
+without any indication that the stack was never applied, so a caller following
+SKILL.md's "never assume a stack" guidance could believe stack guidance was
+part of the generated design system. The combination must stay valid, but it
+must say that --stack was ignored.
+
+Stdlib-only (unittest, not pytest) to match test_core.py -- this project ships
+with zero external dependencies.
+
+Run with:
+    python -m unittest discover -s scripts/tests -v
+or directly:
+    python scripts/tests/test_design_system_stack.py
+"""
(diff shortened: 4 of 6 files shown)


clark-cant (approved):
## Review: PR #487 — fix(search): warn that --stack is ignored in --design-system mode

**Verdict: APPROVED** ✅

### Correctness
- The fix is minimal and correct: when `--design-system` and `--stack` are both passed, a one-line note is emitted to stderr before `generate_design_system()` runs. The note clearly states `--stack` is ignored and suggests running a separate `--stack` query.
- The warning goes to stderr, keeping stdout clean for `--json` consumers — correct choice.
- The combination stays valid (exit 0), which is the non-breaking option from the triage discussion.

### Asset sync
- Fix applied in `src/ui-ux-pro-max/scripts/search.py` (source of truth) and mirrored to `.claude/skills/` and `cli/assets/` — all three copies are consistent.
- Test file `test_design_system_stack.py` is mirrored to all three locations as well.

### Test coverage
- Three test cases cover the exact scenarios: (1) both flags → warns on stderr, (2) design-system alone → quiet, (3) stack alone → quiet.
- Uses stdlib `unittest` with subprocess end-to-end invocation — matches project convention (`test_core.py`).
- Verified red without fix, green with fix.

### Checklist
- [x] Source-of-truth changes in `src/`
- [x] Asset sync verified
- [x] Tests added (stdlib, no new deps)
- [x] Conventional commit format
- [x] Feature branch, not main

### Notes
- This directly addresses Finding 1 (priority-1 quick win) from issue #484.
- Findings 2-4 remain open and are correctly out of scope for this PR.

*Reviewed by github-maintain automation*

PR #498, opened 2026-09-15
fix(design): make workflows self-contained, add bundled-skill contract (#474 finding 2)

notbucki:
## Summary

Finding 2 of #474, per the maintainer decision there (2026-09-15): the `design` skill's banner and social-photos workflows named skills this plugin does not ship — `frontend-design`, `ai-artist`, `ai-multimodal`, `chrome-devtools`, `project-management`, `assets-organizing` — and claudekit command namespaces (`/ckm:brand`, `/ckm:design-system`, `/ck:ui-ux-pro-max`, `/ck:frontend-design`). Under a plugin or CLI install those steps fail or leave the agent improvising. This PR makes the skill self-contained after the pattern of #473 and adds the grep contract to `check-asset-sync.yml`.

## What changed

- `.claude/skills/design/SKILL.md` — banner workflow (research / design / export) and social-photos workflow (orchestrate / design / export / verify / organize) rewritten: bundled skills stay referenced by name, optional tooling (browser research, image generation, screenshot capture) becomes an optional runtime capability with a stated fallback, unbundled steps are implemented inline or dropped. `Related Skills` keeps only bundled names; the routing table and the Integration section say "Bundled" instead of "External" for `brand` / `design-system` / `ui-styling`, which is what they are under both installs.
- `.claude/skills/design/references/social-photos-design.md` — the same class of references one directory down (Step 1 `project-management`, Step 4 `/ckm:` and `/ck:` commands, Step 5 / 6 `chrome-devtools` and Chrome MCP, Step 8 `assets-organizing`, the scope note's `ai-artist`), rewritten the same way. The #476 → #482 history of this issue is why the fix covers the skill directory, not only `SKILL.md`.
- `SKILL.md` and `references/logo-design.md` — the bare `/ui-ux-pro-max` command (two lines) becomes "use the bundled `ui-ux-pro-max` skill": the bare slash form is the CLI / manual-install spelling, while the plugin registers the skill as `ui-ux-pro-max:ui-ux-pro-max`. Same rationale, one step beyond the decision's line list — easy to drop if you would rather keep it separate.
- `cli/assets/skills/design/…` — the three files mirrored, byte-identical to the source copies (what `sync-assets.mjs` produces for them; a full `sync:assets` run would also rewrite the two `brand` files noted under Testing, so the mirror was done per file).
[...]

PR #497, opened 2026-09-14
fix(brand): stop adjustBrightness collapsing dark color shades to black

nadiiasoshenko-netizen:
## Summary

`adjustBrightness()` in `cli/assets/skills/brand/scripts/sync-brand-to-tokens.cjs` generates the 700/800/900 shades of each color scale by adding/subtracting a flat `255 * percent` to every RGB channel. For a **dark** base color (channels already close to 0), darkening by `-0.3`/`-0.45`/`-0.6` clamps every channel to 0 — so shades 700, 800, and 900 all come back as the identical `#000000` instead of a graded dark scale.

This isn't an edge case: any brand whose primary color is a dark tone (common in coffee, luxury, editorial, and dark-mode-first brands) hits it. Repro with a coffee-roastery-style primary color `#4A3228`:

```
$ node sync-brand-to-tokens.cjs   # docs/brand-guidelines.md has Primary Color = #4A3228
...
primitive.color.primary.700 -> #000000
primitive.color.primary.800 -> #000000
primitive.color.primary.900 -> #000000
```

## Fix

Blend each channel proportionally toward white (`percent > 0`) or toward black (`percent < 0`) instead of a flat additive shift, so the amount of change scales with how much headroom the channel actually has. After the fix, the same input produces a proper graded scale:

```
700 -> #34231C
800 -> #291C16
900 -> #1E1410
```

Verified this doesn't regress the light end or existing lighter/mid-tone brand colors (checked against the bundled starter template's blue/purple/green palette, and a synthetic pure-black / pure-white / mid-gray base).

## Test plan

- [x] Added `test_dark_base_color_does_not_collapse_shades_to_black` to `test_sync_brand_to_tokens.py`, which syncs a dark brand color and asserts the 700/800/900 shades stay distinct and non-black
- [x] `python3 -m pytest cli/assets/skills/brand/scripts/tests/test_sync_brand_to_tokens.py -v` — all 3 tests pass (including the pre-existing two)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MmtTPnoqMrjjhKjod9VTzr

PR #499, opened 2026-09-15
fix(brand): apply #497 to the source copy so check:assets passes again

notbucki:
## Summary

`npm run check:assets` has failed on `main` since #497: that PR patched `cli/assets/skills/brand/scripts/sync-brand-to-tokens.cjs` and its test, but `sync-assets.mjs` treats `.claude/skills/` as the source of truth and mirrors it into `cli/assets/`, so the source copy still carries the old `adjustBrightness` and the sync check reports two stale asset files. Because that check is the first step of both the "Check asset sync" and the "Tests" workflows, every PR since then stops there and skips the path contract and the regression tests (#498 included).

## What changed

The same two files, copied from `cli/assets/skills/brand/scripts/` to `.claude/skills/brand/scripts/`. The diff of this PR is byte-identical to the diff #497 applied to the CLI copy (compared hunk by hunk); nothing else. Running `sync:assets` instead would have gone the other way and reverted #497.

## Testing

- `npm --prefix cli run check:assets`: "Assets are in sync."
- `python3 -m pytest .claude/skills/brand/scripts/tests/test_sync_brand_to_tokens.py`: 3 passed — the test #497 added, now against the source copy.
- `python3 -m pytest .claude/skills`: 235 passed.
- `git diff --check` clean.

If you would rather sync this yourself, feel free to close.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Claude Fable 5.1), which prepared the copy and this description; a human reviewed and approved the submission.


clark-cant (approved):
Approved — clean source-of-truth sync fix.

**Summary:** Applies the same `adjustBrightness` fix from PR #497 to the `.claude/skills/` source copy, restoring `check:assets` on main.

**Mandatory gates:**
- Duplicate/prior implementation: clear — this is the source-of-truth counterpart to the CLI-side fix in #497
- Project standards: clear — follows established skill-relative path convention
- Strategic necessity: clear — main is broken without this (check:assets fails, blocking all PRs)

**Review:**
- Diff is byte-identical to the #497 fix applied to the source copy ✅
- New test (`test_dark_base_color_does_not_collapse_shades_to_black`) correctly validates the fix against a dark brand color ✅
- CI green: Check asset sync SUCCESS, Tests SUCCESS ✅
- mergeStateStatus: CLEAN ✅

Merging now to unblock main.

*Posted by github-maintain cron*

PR #434, opened 2026-07-28
fix(design-system): dark style no longer returns a light palette + contradictory anti-pattern (#428)

noron12234:
Fixes #428.

## Root cause

`DesignSystemGenerator.generate()` ranks each domain independently:

```python
best_style = self._select_best_match(style_results, effective_style_priority)  # mode-aware
best_color = color_results[0] if color_results else {}                         # top hit, no mode filter
...
"anti_patterns": reasoning.get("anti_patterns", ""),                           # category rule, no mode gate
```

Only the style resolution knows anything about light vs dark, so the three can
disagree. Reproduced on the issue's exact command:

```console
$ python3 scripts/search.py "SaaS invoicing fintech B2B professional dark mode" --design-system -f markdown

### Style
- **Name:** Modern Dark (Cinema Mobile)
- **Mode Support:** Light ✓ Light mode only as exception | Dark ✓ Dark Mode Primary
### Colors
| Background | `#F8FAFC` |   ← light
| Foreground | `#020617` |
### Avoid (Anti-patterns)
- Dark mode by default        ← contradicts the style it just recommended
```

## Fix

Resolve the mode once, then make the palette and anti-patterns answer to it:

1. `_resolve_color_mode(query, style)` — dark when the query asks for it
   (`dark mode`, `dark theme`, `oled`, …) **or** the style's `Light Mode ✓` /
   `Dark Mode ✓` columns mark it dark-first (`Dark Mode Primary`, `Dark-only`,
   `Light mode only as exception`, …). `✓ Full` on both, which is what 53 of the
   87 styles carry, stays light unless the query says otherwise.
2. `_select_palette_for_mode()` — in dark mode, take the highest-ranked palette
   whose `Background` is a dark surface (WCAG relative luminance < 0.18).
3. `_filter_anti_patterns_for_mode()` — drop `Dark mode by default`-style clauses
   from the `+`-joined anti-pattern string once dark is resolved.

`colors.csv` has no mode column, so the classifier reads `Background` luminance.
The data separates cleanly: the lightest dark background (`#1F2937`) sits at
~0.026 and the darkest light background (`#E8ECF1`) at ~0.79.

## Deliberately narrow

- **Only the dark case filters.** Light keeps the existing "top hit wins"
  behaviour, so any query that never mentions a mode produces byte-identical
  output. Making it symmetric would repaint a lot of existing results and felt
  like a separate call for you to make.
- **Dark falls back to the top hit** when `colors.csv` has no dark ramp for that
[...]

Issue #428, opened 2026-07-21
--design-system returns a light palette for a dark style, plus a contradictory anti-pattern


amazscript:
## Summary

For a dark-mode query, `--design-system` returns three mutually contradictory pieces of advice in a single output: a **dark** style, a **light** palette, and an anti-pattern telling you not to use dark mode.

## Reproduction

Run against `main` (commit `5c0946f`), from `src/ui-ux-pro-max/`:

```bash
python3 scripts/search.py "SaaS invoicing fintech B2B professional dark mode" --design-system -f markdown
```

Relevant lines of the output:

```
### Style
- **Name:** Modern Dark (Cinema Mobile)
- **Mode Support:** Light ✓ Light mode only as exception | Dark ✓ Dark Mode Primary

### Colors
| Background | `#F8FAFC` | `--color-background` |
| Foreground | `#020617` | `--color-foreground` |

### Avoid (Anti-patterns)
- Dark mode by default
```

## Expected

When the resolved style is `Dark Mode Primary`, the palette should be its dark ramp — `Background` should be a dark surface, not `#F8FAFC`. And the generic "Dark mode by default" anti-pattern should be suppressed when the query explicitly asks for dark mode, or when the selected style is dark-primary.

## Actual

Style, palette and anti-patterns appear to be resolved independently, so they can disagree. Following the output literally produces near-invisible text: `Foreground #020617` on `Background #F8FAFC` is a light theme, while every effect and style keyword describes a dark one.

## Impact

The palette is the part users copy verbatim into CSS variables. An agent that trusts the output ships a light theme with dark-theme styling instructions attached; one that notices the conflict has to rebuild the dark ramp by hand and re-verify contrast, which is exactly the work the design system is supposed to remove.

## Suggestion

Filter palette candidates by the resolved style's mode support before ranking them, and gate mode-related anti-patterns on the resolved mode and the query keywords.

## Related — a merged fix that is not yet on npm

While tracing this I noticed #411 ("docs: sync --domain and --stack lists with search.py --help", merged 2026-07-13) landed **after** the `v2.11.0` tag:

```
git merge-base --is-ancestor 0468898 v2.11.0  →  false
```

[...]

PR #500, opened 2026-09-16
fix(design-system): agree on dark mode in light results, keep box output in bounds, readable button text

alinultimatltd:
## What & why

Found by running the README example, `search.py "beauty spa wellness" --design-system -p "Serenity Spa"`. There are three output bugs, all in `design_system.py`.

### 1. Light results contradict themselves on dark mode
#434 fixed the dark side of #428. On the light side, the STYLE block still prints `Mode Support: Light supported  Dark supported` while AVOID says `... + Dark mode`.

`_mode_support_labels()` now qualifies the dark label when the anti-patterns (after `_filter_anti_patterns_for_mode`) still contain a dark-mode clause:

```
Mode Support: Light supported | Dark supported, but avoid for this product
```

This applies to the ASCII and markdown output. `not-recommended` is left as is.

### 2. ASCII box lines overflow the border
- Conversion, CTA, Performance/Accessibility, the Google Fonts URL and motion Trigger were appended as single lines. They now go through `wrap_text`, which also splits tokens longer than a line, such as URLs.
- Every `section_header` was called with `BOX_WIDTH + 1`, which made it one character wider than the box. The CSS Import truncation was also one character too long.
- The colour label column was 14 wide, which misaligned `Card Foreground:`, `Muted Foreground:` and `On Destructive:`. It is now 18.

I checked all 192 `ui-reasoning.csv` categories, with and without `--motion/--variance/--density 8`. Every line now measures exactly `BOX_WIDTH + 1`.

### 3. MASTER.md button specs ignore the palette's contrast pairs
- `.btn-primary` hard-coded `color: white`, which is below 4.5:1 on the palette's Accent in **137 / 192** palettes. It now uses `On Accent`, which every row defines.
- `.btn-secondary` uses Primary as text on a transparent button, which is below 4.5:1 against Background in **94 / 192** palettes. It now falls back to Foreground in those cases. The border stays Primary.

Serenity Spa: primary button text 4.23:1 → 4.96:1, secondary button text 3.23:1 → 8.84:1.

## Tests
- Added `TestLightResultOutputCoherence` to `test_design_system_mode.py`: dark-label qualification, an unchanged label when dark isn't avoided, exact box width, and 4.5:1 for both button specs in MASTER.md.
- `npm --prefix cli run check:assets` passes, with the three copies synced.
- `python3 -m pytest .claude/skills` passes: 239 tests.
- `python3 -m unittest discover -s tests` passes in all three script trees: 164 tests.


clark-cant (approved):
## Maintainer Review — Approve

**Summary:** Fixes three output bugs in `design_system.py`: (1) light-mode results no longer print "Dark supported" next to an AVOID clause that says "Dark mode"; (2) the ASCII box no longer overflows its border (long fields wrap, over-long tokens split, section headers were 1 char too wide, colour label column widened to 18); (3) `MASTER.md` button specs use the palette's On Accent / Foreground instead of hard-coded white / Primary when the pair fails 4.5:1.

**Risk level:** Low

### Mandatory gates

- **Duplicate / prior implementation:** clear — this is the light-side counterpart of the merged #434 (dark side of #428). No other open PR or issue covers box-width overflow or button contrast.
- **Project standards:** clear — edits the source of truth under `src/ui-ux-pro-max/` and mirrors it into `cli/assets/` and `.claude/skills/`. All six changed files are byte-identical across the three copies (md5 verified), so `check:assets` will stay green.
- **Strategic necessity:** clear value — removes self-contradicting guidance from generated output and fixes sub-4.5:1 button text in generated `MASTER.md`.

### Verified independently (not taken from the description)

- Head `79b64af0`, `src/ui-ux-pro-max/scripts`: `python3 -m unittest discover -s tests` → **Ran 164 tests, OK**, including the four new `TestLightResultOutputCoherence` cases.
- `colors.csv` (192 rows): every row defines `On Accent`; white on Accent is below 4.5:1 in exactly **137/192** rows; `On Accent` on Accent meets ≥4.5:1 in **192/192** (the new `.btn-primary` colour is safe); Primary on Background is below 4.5:1 in exactly **94/192**, and the `Foreground` fallback passes in every case.
- `styles.csv` `Dark Mode ✓` vocabulary is `supported | conditional | not-recommended`, so the new `!= "not-recommended"` guard is meaningful.
- Reproduced the motivating case: `DesignSystemGenerator().generate("beauty spa wellness", "Serenity Spa")` → `anti_patterns = "Bright neon colors + Harsh animations + Dark mode"`, `dark_mode = "supported"` — exactly the contradiction being fixed.
- Border arithmetic: `section_header(name, BOX_WIDTH)` now emits exactly `BOX_WIDTH + 1` visible chars, matching every other row and the `╔═╗` / `┌─┐` frame (`w = BOX_WIDTH - 1`).

### Findings

- **Critical:** none
- **Important:** none
- **Suggestion** (non-blocking):
  1. Wrapped Google Fonts / CSS URLs are split mid-token, so a copied line is no longer a complete URL. Reasonable trade-off versus the overflow it fixes, but worth a code comment.
  2. The `[:67]` CSS-Import truncation is a hard-coded width derived from the current `│     CSS Import: ` prefix; deriving it from `BOX_WIDTH` would survive future label changes.
  3. `test_master_button_text_meets_contrast_on_its_background` will raise `AttributeError`/`TypeError` on a regex miss instead of failing cleanly — add `assertIsNotNone(...)` before `.group()`.

### Verdict: Approve

No critical or important issues. Test suite passes, the three mirrored copies are in sync, and the stated numbers all reproduce.

### Merge status: not merged

CI has not run: all three `pull_request` workflow runs for head `79b64af0` ("Check asset sync" and "Tests") are held with `conclusion=action_required` because the contributor is a first-time fork author, and `mergeStateStatus=UNSTABLE`. A human maintainer needs to approve those workflow runs before CI can confirm this change and it can be merged. Merge is deferred to the human maintainer.

*Posted by github-maintain cron at 2026-09-16T11:57:20Z*

